PR #4, opened 2025-11-22
WDY-552: Add WendyOS development container registry

mihai-chiorean:
Implements a development container registry for WendyOS that enables local container image caching and distribution during development.

## Implementation

**Container Registry Image (Pre-built):**
- Downloads pre-built ARM64 registry image from GitHub releases (v1.0.0)
- Uses SHA256 checksum verification for security
- Installed to /usr/share/edgeos/offline-images/
- Recipe: wendyos-dev-registry-image_1.0.0.bb

**Container Registry Service:**
- Systemd service for running the registry container
- Uses nerdctl to run container from containerd's content store
- Type=oneshot with RemainAfterExit for proper container lifecycle
- Listens on 0.0.0.0:5000 (configurable via LISTEN_ADDRESS)
- Disabled by default, started on-demand by wendy-agent
- Recipe: wendyos-containerd-registry_git.bb

**Import Service:**
- One-shot systemd service that imports registry image on first boot
- Uses ctr to import from /usr/share/edgeos/offline-images/
- Creates state file to prevent re-import
- Enabled by default

## Build Integration

- Container image downloaded from GitHub releases, not built during Yocto build
- BitBake automatically unpacks tar.gz to tar for image import
- Added to edgeos-image via wendyos-containerd-registry and wendyos-dev-registry-image packages
- No Go compilation required (binary provided in container image)

## Configuration

Registry container configured via environment variables:
- LISTEN_ADDRESS=0.0.0.0:5000
- CONTAINERD_NAMESPACE=default
- LOG_FORMAT=json
- Logs to journald with tag=wendyos-dev-registry

## Files Changed

- .gitignore: Added *.log pattern
- recipes-containers/wendyos-dev-registry-image/: New recipe for registry image
- recipes-devtools/wendyos-containerd-registry/: New recipe and service files
- recipes-core/images/edgeos-image.bb: Added dev-registry packages

### DIFF
--- a/recipes-containers/wendyos-dev-registry-image/wendyos-dev-registry-image_1.0.0.bb
+++ b/recipes-containers/wendyos-dev-registry-image/wendyos-dev-registry-image_1.0.0.bb
@@ -3,14 +3,14 @@ DESCRIPTION = "Pre-built container image for the WendyOS development registry"
 LICENSE = "MIT"
 LIC_FILES_CHKSUM = "file://${COMMON_LICENSE_DIR}/MIT;md5=0835ade698e0bcf8506ecda2f7b4f302"
 
-PV = "1.0.0"
+PV = "1.0.7"
 
 # Download the container image tar from GitHub releases
-# BitBake will automatically unpack the .tar.gz to get the .tar file
-SRC_URI = "https://github.com/wendylabsinc/containerd-registry/releases/download/v${PV}/containerd-registry-arm64.tar.gz"
+# Note: Using ;unpack=0 because the artifact format requires manual handling
+SRC_URI = "https://github.com/wendylabsinc/containerd-registry/releases/download/v${PV}/containerd-registry-arm64.tar.gz;unpack=0"
 
-# Checksum for v1.0.0 release
-SRC_URI[sha256sum] = "0ea9497b4fd3b6ed3a2b61bae71d27c595c058a72d76621379ef56fe6c8b5073"
+# Checksum for v1.0.7 release
+SRC_URI[sha256sum] = "af3b5a919acf2ff799b19b3fab16ca61f05f6b7192b4b16aece6a3cc0723f828"
 
 inherit allarch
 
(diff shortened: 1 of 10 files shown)


recipes-containers/wendyos-dev-registry-image/wendyos-dev-registry-image_1.0.0.bb
```diff
@@ -0,0 +1,34 @@
+SUMMARY = "WendyOS Development Container Registry Image"
+DESCRIPTION = "Pre-built container image for the WendyOS development registry"
+LICENSE = "MIT"
+LIC_FILES_CHKSUM = "file://${COMMON_LICENSE_DIR}/MIT;md5=0835ade698e0bcf8506ecda2f7b4f302"
+
+PV = "1.0.0"
+
+# Download the container image tar from GitHub releases
+# BitBake will automatically unpack the .tar.gz to get the .tar file
+SRC_URI = "https://github.com/mihai-chiorean/containerd-registry/releases/download/v${PV}/containerd-registry-arm64.tar.gz"
```
Joannis: Can we create a github repo under the wendylabsinc org for pre-built images? This idea has come up more often, and it would be good to put it in a central location under our org. Maybe we can use GHCR or some other public container registry for this? Google cloud also has one, and I think every cloud does.

recipes-devtools/wendyos-containerd-registry/files/edgeos-dev-registry-import.service
```diff
@@ -0,0 +1,25 @@
+[Unit]
+Description=EdgeOS Dev Registry Image Import (First Boot)
```
Joannis: Why is this new script named EdgeOS and the rest WendyOS? Oversight or with some purpose?

recipes-containers/wendyos-dev-registry-image/wendyos-dev-registry-image_1.0.0.bb
mihai-chiorean: Yep. I'll move the repo. For the registry though in this particular case, for the way we build the os, just downloading a tar.gz is really smooth. Yocto actually auto detects gz and unzips it. I wouldn't bother setting up ghcr just for this. It doesn't help the use case.

recipes-devtools/wendyos-containerd-registry/files/edgeos-dev-registry-import.service
mihai-chiorean: Oversight

recipes-devtools/wendyos-containerd-registry/files/wendyos-dev-registry.sh
```diff
@@ -0,0 +1,186 @@
...
+
+restart_registry() {
+    stop_registry
+    sleep 1
+    start_registry
+}
+
+# Main command dispatch
+COMMAND="${1:-}"
+CUSTOM_LISTEN_ADDRESS="${2:-}"
```
abularca, line 159: If I understand correctly, the listening address can be provided through the **LISTEN_ADDRESS** environment variable, the script positional argument or both, which is a bit confusing.

abularca (approved):

PR #5, opened 2025-11-23
Add USB network performance tuning with kernel parameters

mihai-chiorean:
Implements kernel-level network optimizations for USB gadget networking to improve throughput and reduce latency.

Key optimizations:
- Increased socket buffer sizes (16MB max, 2MB default)
- Tuned TCP buffer sizes for better throughput
- BBR congestion control for variable latency links
- Increased network device backlog and budget
- Enabled TCP fast open and optimized timeouts
- Enabled TCP SACK and timestamps

The tuning is applied via sysctl.d configuration and automatically loaded by systemd-sysctl on boot.

Only enabled when EDGEOS_USB_GADGET=1 is set.

Resolves: WDY-542

### DIFF
--- a/recipes-support/usb-network-tuning/files/99-usb-network-performance.conf
+++ b/recipes-support/usb-network-tuning/files/99-usb-network-performance.conf
@@ -20,6 +20,9 @@ net.core.netdev_max_backlog = 5000
 # Enable TCP window scaling for better throughput on high-latency links
 net.ipv4.tcp_window_scaling = 1
 
+# Enable fq packet scheduler (required for BBR pacing)
+net.core.default_qdisc = fq
+
 # Use BBR congestion control for better performance over variable latency
 # Falls back to cubic if BBR is not available
 net.ipv4.tcp_congestion_control = bbr
--- a/recipes-support/usb-network-tuning/files/99-usb-network-performance.conf
+++ b/recipes-support/usb-network-tuning/files/99-usb-network-performance.conf
@@ -1,0 +1,40 @@
+# USB Network Performance Tuning
+# Optimizes kernel parameters for better USB gadget network throughput
+
+# Increase maximum socket buffer sizes (16MB)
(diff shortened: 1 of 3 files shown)


recipes-support/usb-network-tuning/files/99-usb-network-performance.conf
```diff
@@ -0,0 +1,40 @@
...
+
+# Increase network device backlog queue length
+net.core.netdev_max_backlog = 5000
+
+# Enable TCP window scaling for better throughput on high-latency links
+net.ipv4.tcp_window_scaling = 1
+
+# Use BBR congestion control for better performance over variable latency
+# Falls back to cubic if BBR is not available
+net.ipv4.tcp_congestion_control = bbr
```
abularca, line 28: In theory at least, the **bbr** congestion algorithm works better paired with **fq** as packet scheduler (qdisc) so, this would probably help (or at least it worth a try):
```
# Enable fq packet scheduler (required for BBR pacing)
net.core.default_qdisc = fq

# Use BBR congestion control
net.ipv4.tcp_congestion_control = bbr
```

recipes-support/usb-network-tuning/files/99-usb-network-performance.conf
```diff
@@ -0,0 +1,40 @@
...
+# Increase maximum socket buffer sizes (16MB)
+net.core.rmem_max = 16777216
+net.core.wmem_max = 16777216
+
+# Increase default socket buffer sizes (2MB)
+net.core.rmem_default = 2097152
+net.core.wmem_default = 2097152
+
+# TCP buffer sizes: min, default, max (in bytes)
+# min: 4KB, default: 87KB, max: 16MB
```
abularca, line 13: Large buffers potentially increase memory usage and buffering, which can lead to latency increase.
Would it make sense to reduce the maximum socket buffer maybe to 8MB?

PR #22, opened 2026-01-23
Add conditional boot chain switching for atomic UEFI capsule updates

abularca:
Implements smart boot chain management to support both capsule-driven
and OS-driven update flows, following NVIDIA's boot chain architecture
design.
The switch-rootfs script pre-checks for bootloader update markers before
deciding how to switch boot chains:

1. Mount inactive rootfs partition
2. Check for /var/lib/edgeos/update-bootloader marker
3. Based on marker presence:

 CAPSULE UPDATE (marker exists):
 - Skip nvbootctrl call entirely
 - Stage capsule to ESP /EFI/UpdateCapsule/
 - Set OSIndications EFI variable (0x04)
 - UEFI firmware processes capsule on next boot:
   * Writes bootloader to inactive boot chain
   * Calls UpdateFwChain() to update BCT
   * Switches boot chain atomically
 - L4TLauncher reads boot chain from BCT and boots corresponding rootfs
 - Result: Atomic bootloader + rootfs update

 ROOTFS-ONLY UPDATE (no marker):
 - Call nvbootctrl set-active-boot-slot
 - Sets BootChainFwNext EFI variable
 - HandleBootChainUpdate() processes BC_NEXT on next boot
 - Switches boot chain via UpdateFwChain()
 - Result: Rootfs-only update, bootloader unchanged

KEY DESIGN:
- Rootfs slot is always linked with bootloader chain (Boot Chain 0 -> Rootfs Slot 0, Boot Chain 1 -> Rootfs Slot 1)
- Capsule update handles boot chain switching via UpdateFwChain() in UEFI firmware
- No manual nvbootctrl needed for capsule updates (prevents BC_NEXT conflict with CheckAndCancelUpdate())
- Maintains backward compatibility for rootfs-only updates


recipes-bsp/uefi/files/UefiDefaultSecurityKeys.dts
mihai-chiorean, line 1: are these ok? do we need to generate newer ones?

recipes-mender/tegra-state-scripts/files/switch-rootfs
```diff
@@ -110,23 +130,74 @@ esac
...
+#    - System reboots to new boot chain with new rootfs
+###############################################################################
+
+# Check if bootloader update capsule will be staged
+marker_rel="/var/lib/edgeos/update-bootloader"
+mount_dir="/mnt/inactive"
+will_stage_capsule=0
+
+mkdir -p "$mount_dir"
+if mount "${inactive_part}" "${mount_dir}" 2>/dev/null
```
mihai-chiorean, line 166: does this risk a race condition with `maybe_stage_capsule()` below? looks like they both call mount

uefi-keys/generate-keys.sh
```diff
@@ -0,0 +1,237 @@
...
+  "${GITHUB_URL}|TestRoot.pub.pem|${KEY_EXCHANGE_KEY}"
+  "${GITHUB_URL}|TestSub.pub.pem|${DB_KEY}"
+)
+
+download() {
+  local url="$1"
+  local output="$2"
+
+  # -q: quiet, --show-progress: progress bar, -O: output file
+  wget -q --show-progress -O "${output}" "${url}"
```
mihai-chiorean, line 44: does this need some error handling if wget fails? maybe `command -v`?

recipes-mender/tegra-state-scripts/files/switch-rootfs
```diff
@@ -110,23 +130,74 @@ esac
...
+        will_stage_capsule=1
+        log "bootloader update marker found: will stage capsule"
+    fi
+    umount "${mount_dir}" || true
+else
+    log "WARN: could not mount inactive rootfs (${inactive_part})"
+fi
+
+# Switch boot chain based on update type
+if [ "${will_stage_capsule}" -eq 1 ]
```
mihai-chiorean, line 179: this determines if we should run nvbootctrl, but then there's a second check in  maybe_stage_capsule(). do we need both?

also, the partition is mounted, checked, unmounted, then mounted again in maybe_stage_capsule(). do we need this cycle? introduces a window for something to happen.

mihai-chiorean: the decision to skip nvbootctrl depends on the first check, but the actual staging depends on the second check. If they ever disagree (e.g., marker deleted between checks), you'd get inconsistent behavior

recipes-mender/tegra-state-scripts/files/verify-bootloader-update
```diff
@@ -0,0 +1,204 @@
...
+else
+    log "WARNING: no pre-update version recorded, cannot compare versions"
+    # Continue to ESRT check
+fi
+
+###############################################################################
+# SECONDARY VERIFICATION: ESRT Status (NVIDIA Official Recommendation)
+###############################################################################
+
+ESRT_STATUS_FILE="/sys/firmware/efi/esrt/entries/entry0/last_attempt_status"
```
mihai-chiorean, line 78: that entry0, is that always the correct one? can it be different?

recipes-mender/tegra-state-scripts/files/switch-rootfs
```diff
@@ -188,21 +272,43 @@ maybe_stage_capsule() {
...
+                umount "${mount_dir}" || true
+                exit 1
+            }
+
+            # Verify OSIndications was written correctly (bit 2 set for capsule processing)
+            # Per UEFI spec, SetVariable() is synchronous - write completes before return
+            osind_file="/sys/firmware/efi/efivars/OsIndications-8be4df61-93ca-11d2-aa0d-00e098032b8c"
+            if [ -r "$osind_file" ]; then
+                # Read byte at offset 4 (skip 4-byte attribute header)
+                value=$(dd if="$osind_file" bs=1 skip=4 count=1 2>/dev/null | od -An -tu1 | tr -d ' ')
```
mihai-chiorean, line 286: would this be more robust with `efivar`? I think that tool might be availabe

mihai-chiorean (approved):

recipes-bsp/uefi/files/UefiDefaultSecurityKeys.dts
abularca: They are the development keys. For production, we need to generate different ones.
At the moment, we don't use secure boot, this is something we need to address as well. I added these files to not loose the signing support but we will have to (re)enable it at some point.

PR #26, opened 2026-02-03
Fix internet sharing

Joannis:
This seems to have fixed my internet sharing. I get an IP address etc even when just connected over USB. I can also use WiFi on the side.


abularca:
Indeed, this seems to be a better approach.

**Before** (method=shared):
- Jetson acts as DHCP server using NetworkManager's built-in connection sharing
- Static IP: 10.42.0.1/24
- NetworkManager automatically spawns dnsmasq for DHCP service

**After** (method=auto):
- Jetson acts as DHCP client
- Expects to receive IP from host computer's internet sharing service

Regarding the `10-usb-gadget-shared.conf` file:

```
[ipv4]
  # DHCP range for connection sharing
  # NetworkManager's dnsmasq will serve DHCP on shared connections
  dhcp-timeout=300
```

The file name and content are misleading for the new configuration.
These settings are more relevant to `method=shared` but not to `method=auto`.

**TODOs**:
1. Rename the file to `10-usb-gadget.conf` and update its content to reflects the topology change
2. Remove dnsmasq dependency from networkmanager_%.bbappend as it is not needed
   (Jetson is DHCP client now)
3. Update comments in the .bbappend file to reflect the new approach

mihai-chiorean (approved):

PR #27, opened 2026-02-03
Rename 'edge' to 'wendy'

abularca:
## Summary

Change `edge` to `wendy` across the entire meta layer, both in file names and file content.
This is a complete re-brand from EdgeOS to WendyOS with no migration path for OTA updates.
Existing EdgeOS devices require complete re-flash.
User data migration requires manual backup/restore.


mihai-chiorean (approved):
looks good I fixed a typo

PR #36, opened 2026-03-24
Wendy Config Partition — Jetson

konstantinbe:
- Issue: [Implement WENDYCONFIG partition on Jetson (WDY-777)](https://linear.app/wendylabsinc/issue/WDY-777/implement-wendyconfig-partition-on-jetson)
- Integration PR: https://github.com/wendylabsinc/WendyOS/pull/35


konstantinbe:
https://linear.app/wendylabsinc/issue/WDY-777/implement-wendyconfig-partition-on-jetson

meta-tegra-extensions/recipes-bsp/wendy-config-partition/wendy-config-partition_1.0.bb
```diff
@@ -0,0 +1,28 @@
+SUMMARY = "WendyOS first-boot config partition (FAT32, 64 MB)"
+DESCRIPTION = "Creates a 64 MB FAT32 image labeled WENDYCONFIG for use as \
```
abularca, line 2: The description says for use on NVMe storage but COMPATIBLE_MACHINE includes the SD card machine.

meta-tegra-extensions/recipes-bsp/wendy-config-partition/wendy-config-partition_1.0.bb
```diff
@@ -0,0 +1,28 @@
...
+the first-boot configuration partition on NVMe storage. \
+Auto-mounts as /Volumes/WENDYCONFIG on macOS."
+
+LICENSE = "MIT"
+LIC_FILES_CHKSUM = "file://${COMMON_LICENSE_DIR}/MIT;md5=0835ade698e0bcf8506ecda2f7b4f302"
+
+DEPENDS = "dosfstools-native"
+
+COMPATIBLE_MACHINE = "(jetson-orin-nano-devkit-nvme-wendyos|jetson-agx-orin-devkit-nvme-wendyos|jetson-orin-nano-devkit-wendyos)"
```
abularca, line 11: We might have a consistency issue.
Is this valid for all Tegra machines? If yes, it is probably better to use `(tegra)` instead of narrowing down to specific machines and let the layout bbappend control inclusion.
If a new tegra machine is later added (e.g. jetson-agx-orin-devkit-wendyos), the `:append:tegra` adds wendy-config-partition as a dependency, but COMPATIBLE_MACHINE rejects it therefore, the recipe gets skipped, the dependency is unsatisfied, and the build fails.
The recipe just creates a FAT32 image in DEPLOY_DIR_IMAGE. It's cheap to build, and the actual partition injection is controlled by the storage layout bbappend which is already per-machine.

recipes-core/images/wendyos-image.bbappend
```diff
@@ -1,27 +1,39 @@
...
+    # Copy wendy-config FAT32 image into the tegraflash package directory so
+    # tegraparser_v2 can find it when processing the partition layout XML.
+    # create_tegraflash_pkg never auto-includes files from DEPLOY_DIR_IMAGE
+    # just because they are referenced in a partition layout XML — each file
+    # must be copied explicitly.  Required for both NVMe and SD card builds.
+    if [ -f "${DEPLOY_DIR_IMAGE}/wendy-config.fat32.img" ]; then
+        cp "${DEPLOY_DIR_IMAGE}/wendy-config.fat32.img" ./wendy-config.fat32.img
+        bbnote "Copied wendy-config.fat32.img into tegraflash package"
+    else
+        bberror "wendy-config.fat32.img not found in DEPLOY_DIR_IMAGE (${DEPLOY_DIR_IMAGE})"
```
abularca, line 36: I suppose in case of an error we would like to stop the build process.
`bberror` does not stop the build, it just writes the message and continues.
There are two approaches in this case:
- either use `bbfatal` - stops immediatelly the build
- or keep the `bberror` and return 1 so that the shell function fails

meta-tegra-extensions/recipes-bsp/wendy-config-partition/wendy-config-partition_1.0.bb
```diff
@@ -0,0 +1,28 @@
...
+do_fetch[noexec] = "1"
+do_unpack[noexec] = "1"
+do_patch[noexec] = "1"
+do_configure[noexec] = "1"
+do_compile[noexec] = "1"
+do_install[noexec] = "1"
+
+do_deploy() {
+    dd if=/dev/zero of=${WORKDIR}/wendy-config.fat32.img bs=1M count=64
+    mkfs.fat -F 32 -n "WENDYCONFIG" ${WORKDIR}/wendy-config.fat32.img
```
abularca, line 24: To maintain the same partitions naming, I would go for `config` instead of `WENDYCONFIG`.

meta-tegra-extensions/recipes-bsp/tegra-binaries/tegra-storage-layout-base_%.bbappend
```diff
@@ -1,45 +1,38 @@
...
 
-    bbnote "wendyos: Modifying ${layout_file} to use mender_data partition..."
+            bbnote "wendyos: Modifying ${layout_file} for NVMe (mender_data + wendy_config)..."
 
-    # 1. Remove the "reserved" partition (between UDA and APP)
-    #    This partition blocks expansion and is not needed
-    nvflashxmlparse --remove --partitions-to-remove reserved \
-        --output ${WORKDIR}/${layout_file}.tmp1 \
-        ${layout_path}
+            # 1. Remove the "reserved" partition
```
abularca, line 29: I would keep the original comment, it is more clear

PR #50, opened 2026-04-27
Replace MACHINE/BOARD with DEVICE/STORAGE in board configs and CI

Joannis:
## Summary

- Board `local.conf` templates now use `DEVICE` (matching the publisher identifier, e.g. `raspberry-pi-5`) and `STORAGE` (`sd` or `nvme`) instead of the internal Yocto `MACHINE` name
- New `machine-map.inc` derives the Yocto `MACHINE` from `DEVICE+STORAGE` so BitBake still works transparently
- CI matrices simplified from 4 fields (`device`, `board`, `machine`, `publisher_device`) to 2 (`device`, `storage`); a **Derive build variables** step resolves `BOARD` and `MACHINE` for downstream steps
- QEMU board left unchanged (no sd/nvme concept applies)

## Test plan

- [ ] Verify bootstrap works: `BOARD=rpi5-nvme ./bootstrap.sh`
- [ ] Verify BitBake picks up the correct machine via `machine-map.inc` (check `MACHINE` in `build/conf/local.conf` after bootstrap)
- [ ] Confirm CI build matrix jobs appear as `raspberry-pi-5 (sd)` etc.
- [ ] Confirm publisher `--device` flag still receives `raspberry-pi-5` / `jetson-orin-nano`

### DIFF
--- a/conf/template/include/local/machine-map.inc
+++ b/conf/template/include/local/machine-map.inc
@@ -1,4 +1,0 @@
-# Maps DEVICE + STORAGE to the Yocto MACHINE name.
-# Require this file after setting DEVICE and STORAGE in your board's local.conf.
-
-MACHINE = "${@{'raspberry-pi-3:sd': 'raspberrypi3-64-wendyos', 'raspberry-pi-4:sd': 'raspberrypi4-64-wendyos', 'raspberry-pi-5:sd': 'raspberrypi5-wendyos', 'raspberry-pi-5:nvme': 'raspberrypi5-nvme-wendyos', 'jetson-agx-orin:nvme': 'jetson-agx-orin-devkit-nvme-wendyos', 'jetson-orin-nano:nvme': 'jetson-orin-nano-devkit-nvme-wendyos', 'jetson-orin-nano:sd': 'jetson-orin-nano-devkit-wendyos'}.get((d.getVar('DEVICE') or '') + ':' + (d.getVar('STORAGE') or ''), '')}"
--- a/conf/template/include/local/machine-map.inc
+++ b/conf/template/include/local/machine-map.inc
@@ -1,0 +1,4 @@
+# Maps DEVICE + STORAGE to the Yocto MACHINE name.
+# Require this file after setting DEVICE and STORAGE in your board's local.conf.
+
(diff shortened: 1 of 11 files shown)


abularca:
1. QEMU board template not migrated — `conf/template/boards/qemu-arm64/local.conf`:6, it still uses `MACHINE = "qemuarm64-wendyos"` directly

2. machine-map.inc 
- The mapping is being expressed as runtime computation inside the build system instead of as configuration data alongside the board it applies to (possible evaluation issues)
- Silent fallback on bad input i.e., it could return empty `MACHINE` in case of typos, missing DEVICE/STORAGE, or unsupported combinations
- One-liner of Python embedded in a BitBake string. Hard to read, hard to diff, hard to review
- Mixes a Yocto-internal concept (`MACHINE` selection) with project metadata (DEVICE/STORAGE for publisher / CI). MACHINE is what BitBake consumes while DEVICE/STORAGE belong to the wider system. Putting the resolution inside BitBake gives BitBake authority over a mapping that other tools (Makefiles, scripts, the publisher CLI) might
also want to consume.
- `bitbake -e` is the only way to confirm what MACHINE resolves to, which is slow and requires a build environment
- The lookup table can fall out of sync with reality silently

To keep things clear, I would go for a simpler approach:
Each board's `local.conf` keeps `MACHINE = "..."` explicitly, and also declares DEVICE and STORAGE for publisher/CI metadata. No central mapping, no magic, no silent   fallback:

conf/template/boards/rpi5-sd/local.conf
```
DEVICE  = "raspberry-pi-5"
STORAGE = "sd"
MACHINE = "raspberrypi5-wendyos"
```

At the expense of one additional line, every board file remains self-contained and self-documenting.

Theoretically we could also use an anonymous Python function, but things might get complicated with regards to the order of evaluation. The ${@...} Python expansion in `MACHINE = "..."` must resolve before BitBake selects `machine.conf`.

conf/template/include/local/machine-map.inc
```diff
@@ -0,0 +1,4 @@
+# Maps DEVICE + STORAGE to the Yocto MACHINE name.
+# Require this file after setting DEVICE and STORAGE in your board's local.conf.
+
+MACHINE = "${@{'raspberry-pi-3:sd': 'raspberrypi3-64-wendyos', 'raspberry-pi-4:sd': 'raspberrypi4-64-wendyos', 'raspberry-pi-5:sd': 'raspberrypi5-wendyos', 'raspberry-pi-5:nvme': 'raspberrypi5-nvme-wendyos', 'jetson-agx-orin:nvme': 'jetson-agx-orin-devkit-nvme-wendyos', 'jetson-orin-nano:nvme': 'jetson-orin-nano-devkit-nvme-wendyos', 'jetson-orin-nano:sd': 'jetson-orin-nano-devkit-wendyos'}.get((d.getVar('DEVICE') or '') + ':' + (d.getVar('STORAGE') or ''), '')}"
```
abularca: Please check the generic comment (conversation)

PR #58, opened 2026-04-29
Fix host-build _check-setup and add awscli to AMI deps

justsem:
## Summary

Two bugs surfaced by the first `Build WendyOS Images` run on the new AMI ([run #25133748978](https://github.com/wendylabsinc/WendyOS/actions/runs/25133748978)):

1. **`_check-setup` always ran the Docker check.** The host-build branch and Docker branch lived in two separate `@`-prefixed Make recipe lines, each its own shell. The host-build branch's `exit 0` only ended that shell; Make then proceeded to the Docker check, which failed because no `docker/` dir exists on the AMI (intentionally — host-build skips Docker). Fix is to merge the two into a single `if/else` block in one shell invocation.
2. **`awscli` was missing from the AMI.** `install-build-deps.sh` didn't install it, so the S3 sstate / downloads cache restore + save steps printed `aws: command not found`. The errors are silenced by `|| true`, so the build didn't fail on this — but caching is dead, which means full cold sstate on every run. Adding `awscli` to the apt list fixes it.

The rest of the new pipeline worked as designed: RunsOn picked up our custom AMI (`ami-03768a8d65608ef8f` / `wendyos-builder-2026-04-29-204603`), the bake-time repo prefetch produced `[ok]` for all 8 upstream layers, and bootstrap finished in 24 s.

## Test plan

- [ ] After merge + AMI rebuild, re-run `Build WendyOS Images` workflow on `main`.
- [ ] `Build image` step gets past `_check-setup` and runs `bitbake`.
- [ ] `Restore sstate-cache and downloads from S3 (L2)` and `Save sstate-cache and downloads to S3 (L2)` no longer print `aws: command not found`.

Note: needs the AMI workflow to rebuild after this merges, since `awscli` only lands in the AMI on next bake. The path filter on `scripts/install-build-deps.sh` triggers that automatically.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -503,23 +503,25 @@ _check-machine:
 		exit 1; \
 	fi
 
+# Each @-prefixed recipe line runs in its own shell, so a host-build "exit 0"
+# wouldn't skip subsequent Docker checks. Keep the whole branch in one block.
 _check-setup:
 	@if [ "$(WENDYOS_HOST_BUILD)" = "1" ]; then \
 		if [ ! -d "$(PROJECT_DIR)/repos/poky" ]; then \
 			printf "$(RED)Error: repos/poky not found. Run 'make setup' first.$(NC)\n"; \
 			exit 1; \
 		fi; \
-		exit 0; \
-	fi
-	@if [ ! -d "$(DOCKER_DIR)" ]; then \
-		printf "$(RED)Error: Build environment not set up.$(NC)\n"; \
-		printf "Run 'make setup' first.\n"; \
-		exit 1; \
-	fi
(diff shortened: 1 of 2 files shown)


justsem:
Closing in favor of #59, which is a strict superset of this PR's changes:
- Makefile `_check-setup` host-build branch fix (was here in commit `e432847`)
- AWS CLI v2 install (replacing the apt `awscli` entry that doesn't exist on Noble — was here in commit `a139d52`)

#59 also adds s5cmd, cache-step observability + error propagation, and the Noble unprivileged-userns AppArmor fix. Merging #59 alone gets all of this in one shot.

PR #67, opened 2026-05-05
WDY-977: make OpenSSH server configurable via WENDYOS_SSH

andrei-mazanec:
WDY-977: Added WENDYOS_SSHD option to enable/disable sshd. Default behavior is unchanged (sshd still ships in the image). Set `WENDYOS_SSHD = "0"` in local.conf to drop sshd from the build.

### DIFF
--- a/conf/template/include/local/common.inc
+++ b/conf/template/include/local/common.inc
@@ -24,6 +24,7 @@ CONF_VERSION = "2"
 # USB gadget function ordering (only effective when WENDYOS_USB_GADGET = "1")
 #GADGET_FUNC_ORDER="ecm ncm"
 
-# Debug features - can be overridden in local.conf or machine config
-WENDYOS_DEBUG = "1"
-WENDYOS_DEBUG_UART = "1"
+# Debug/SSH defaults — off for production; opt in via local.conf or board config.
+WENDYOS_DEBUG ?= "0"
+WENDYOS_DEBUG_UART ?= "0"
+WENDYOS_SSHD ?= "0"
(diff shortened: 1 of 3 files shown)


andrei-mazanec:
Regarding the AI Security Review: The debug-tweaks concern is pre-existing and not introduced by this PR — it was already set in the  image recipe before this change.

  The WENDYOS_DEBUG = "1" and WENDYOS_SSHD = "1" defaults are intentional.

  The claim about SSH host key:  debug-tweaks does not touch SSH host key generation. What it does do is set `PermitEmptyPasswords yes`  and `PermitRootLogin yes` in sshd_config, which combined with the empty root
  password is the actual exposure. That's a valid concern for `debug-tweaks`  in general, but again, it predates this PR.

conf/template/include/local/common.inc
```diff
@@ -27,3 +27,6 @@ CONF_VERSION = "2"
 # Debug features - can be overridden in local.conf or machine config
 WENDYOS_DEBUG = "1"
 WENDYOS_DEBUG_UART = "1"
+
+# OpenSSH server (sshd) - on by default; set to "0" to drop sshd from the image.
+WENDYOS_SSHD ?= "1"
```
abularca: We need to check if `WENDYOS_DEBUG`, `WENDYOS_DEBUG_UART` and `WENDYOS_SSHD` have to be disabled by default

recipes-core/images/wendyos-image.bb
```diff
@@ -31,7 +31,6 @@ MENDER_SYSTEMD_AUTO_ENABLE = "1"
 MENDER_CONNECT_ENABLE = "1"
 
 IMAGE_FEATURES += " \
-    ssh-server-openssh \
     debug-tweaks \
```
abularca: `debug-tweaks` should probably be enabled only when `WENDYOS_DEBUG` is enabled

conf/template/include/local/common.inc
andrei-mazanec: Ok. To be discussed in the weekly meeting.

andrei-mazanec:
Joannis confirmed that  WENDYOS_DEBUG, WENDYOS_DEBUG_UART should all be uninstalled/disabled by default. Wendy agent will transfer the artifacts

PR #74, opened 2026-05-11
fix RPi3 boot by switching to MBR disk layout

andrei-mazanec:
The BCM2837 GPU bootrom can only read MBR-partitioned FAT32, not GPT.
  Add rpi-mbr.wks with MBR layout and filesystem labels, a matching fstab
  template, and update cmdline.txt to use root=LABEL=root instead of PARTUUID

### DIFF
--- a/conf/machine/raspberrypi3-64-wendyos.conf
+++ b/conf/machine/raspberrypi3-64-wendyos.conf
@@ -1,11 +1,13 @@
 MACHINEOVERRIDES =. "rpi:raspberrypi3-64:"
 require conf/machine/raspberrypi3-64.conf
 
-# WIC GPT image — no Mender, no tegraflash
+# WIC MBR image — no Mender, no tegraflash. The BCM2837 GPU bootrom on the
+# Pi 3 can only read MBR-partitioned FAT32; the GPT layout used on Pi 4/5
+# (rpi-partuuid.wks) leaves the board unable to find bootcode.bin and the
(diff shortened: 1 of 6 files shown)


andrei-mazanec:
> ## AI Security Review
> This pull request makes targeted build-system changes to support MBR-partitioned SD-card images for the Raspberry Pi 3, replacing a GPT layout that the BCM2837 bootrom cannot read. The changes touch WIC kickstart files, fstab templates, a shell script, and BitBake metadata — none of which handle payment card data, PHI, or direct user authentication flows. The overall risk posture is low. There are no hardcoded secrets, no new network-facing code, and no new third-party dependencies. Two findings warrant attention: a shell-injection vector in the `expand-rootfs.sh` script introduced by unquoted variable expansion of a value sourced from `blkid`, and an informational note about partition-table type detection reliability. All other observations are minor or informational.

PTTYPE is only used in two places: assigned on line 42 and compared   on line 43 inside [[ ]]. It's never passed to a command or used in eval. [[ "$PTTYPE" == "gpt" ]] is safe —  double-quoted left side, [[ ]] doesn't word-split, and blkid -o value -s PTTYPE returns a kernel-normalized token, not   user input.

  The reliability note is already handled: || true on the blkid call means PTTYPE is empty on failure, [[ "" == "gpt" ]]    is false, and sgdisk is skipped.

Joannis (approved):

PR #76, opened 2026-05-13
Add PipeWire V4L2 camera sharing support for headless devices

Joannis:
## Summary

- Adds `60-wireplumber-camera-headless.conf` that marks `monitor.v4l2` as `required` in WirePlumber's main profile, so cameras are enumerated on WendyOS devices that have no logind seat (mirrors the existing Bluetooth headless fix in `50-wireplumber-headless.conf`)
- Adds `pipewire-v4l2` to the image, providing the V4L2 compatibility shim that redirects V4L2-native applications through PipeWire for multi-process camera sharing
- Wires the new config file into the `audio-config` recipe (`SRC_URI`, `do_install`, and `FILES`)

## Test plan

- [ ] Build the image and verify `pipewire-v4l2` is present on the rootfs
- [ ] Confirm `/etc/wireplumber/wireplumber.conf.d/60-wireplumber-camera-headless.conf` is installed
- [ ] Start WirePlumber as the `wendy` user and check `wpctl status` shows the V4L2 camera node
- [ ] Open the camera from two separate processes simultaneously and confirm sharing works without either process receiving a busy/EBUSY error

### DIFF
--- a/recipes-multimedia/audio-config/audio-config_1.0.bb
+++ b/recipes-multimedia/audio-config/audio-config_1.0.bb
@@ -41,7 +41,7 @@ do_install() {
 
     # Install WirePlumber configuration for headless camera (V4L2) support
     # Requires V4L2 monitor so cameras are enumerated without a logind seat
-    install -m 0644 ${WORKDIR}/60-wireplumber-camera-headless.conf \
+    install -m 0644 ${UNPACKDIR}/60-wireplumber-camera-headless.conf \
         ${D}${sysconfdir}/wireplumber/wireplumber.conf.d/
 
     # Install D-Bus policy for Bluetooth access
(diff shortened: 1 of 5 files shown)


recipes-multimedia/audio-config/audio-config_1.0.bb
```diff
@@ -38,6 +39,11 @@ do_install() {
     install -m 0644 ${UNPACKDIR}/50-wireplumber-headless.conf \
         ${D}${sysconfdir}/wireplumber/wireplumber.conf.d/
 
+    # Install WirePlumber configuration for headless camera (V4L2) support
+    # Requires V4L2 monitor so cameras are enumerated without a logind seat
+    install -m 0644 ${WORKDIR}/60-wireplumber-camera-headless.conf \
```
abularca: `WORKDIR` must be replaced with `UNPACKDIR`, otherwise `do_install` will fail on `wrynose`

Joannis: @copilot fix

PR #129, opened 2026-06-16
Bump Jetson rootfs to 16GB (Thor + all Orin boards)

Joannis:
## What

Bump the root filesystem image size from 8GB to **16GB** across all Jetson boards, and move the Orin Nano SD board to a 64GB layout so it can fit.

## Why

`IMAGE_ROOTFS_SIZE ?= "8192"` (8GB) in `recipes-core/images/wendyos-image.bb` is the size of the **rootfs filesystem**, and nothing overrode it for the Jetsons. The per-size Tegra helper (`flash-image-sizes.inc`) only sizes the **device and Mender storage** — not the rootfs — so every Jetson's OS filesystem was capped at 8GB regardless of how big its partitions were.

That's tight as we add features, especially given the NVIDIA CUDA / TensorRT / DeepStream stack (full JetPack on-device is ~16–20GB+). 16GB is nearly free here since the ext4 is sparse and the partitions have ample room.

## Changes

Per-board `IMAGE_ROOTFS_SIZE = "16384"` override on:

| Board | Machine conf | Layout | Notes |
|-------|--------------|--------|-------|
| Thor (AGX Thor NVMe) | `jetson-agx-thor-devkit-nvme-wendyos.conf` | wendy OTA | `flash-image-sizes.inc` is Mender-only, so Thor was on the 8GB default |
| AGX Orin NVMe | `jetson-agx-orin-devkit-nvme-wendyos.conf` | 64GB | ~24GB A/B slots |
| AGX Orin eMMC | `jetson-agx-orin-devkit-emmc-wendyos.conf` | 64GB | ~24GB A/B slots |
| Orin Nano NVMe | `jetson-orin-nano-devkit-nvme-wendyos.conf` | 64GB | ~24GB A/B slots |
| Orin Nano SD | `jetson-orin-nano-devkit-wendyos.conf` | **32GB → 64GB** | bumped layout (see below) |

### Orin Nano SD: 32GB → 64GB layout

The 32GB SD layout (~12.8GB per A/B slot) couldn't hold a 16GB rootfs. Moved it to the 64GB layout — `WENDYOS_FLASH_IMAGE_SIZE = "64GB"` + `TEGRAFLASH_SDCARD_SIZE = "64G"` kept in lockstep — which gives ~24GB slots. **This requires a 64GB minimum SD card** (32GB was too small for practical use anyway); documented in the README SD-card flashing section.

### Not touched

- **QEMU** — sized separately via `QEMU_ROOTFS_SIZE`.
- **Global default / RPi** — left the `8192` default in place rather than bumping globally, to avoid changing RPi (smaller layouts) behavior.

## Verification

Config-only. Recommend a build + flash on each board to confirm the 16GB ext4 fits the derived rootfs partitions end-to-end — particularly the Orin Nano SD on its new 64GB layout, and Thor (the wendy OTA path has no fixed `TEGRA_EXTERNAL_DEVICE_SECTORS`).

### DIFF
--- a/README.md
+++ b/README.md
@@ -320,6 +320,10 @@ build/tmp/deploy/images/<machine>/wendyos-image-<machine>.rootfs.tegraflash.tar.
 
 #### For SD Card Builds
 
+> **Requires a 64 GB (or larger) SD card.** The Orin Nano SD image uses a 64GB
+> layout (`WENDYOS_FLASH_IMAGE_SIZE = "64GB"`) so the A/B rootfs slots can hold
+> the 16GB rootfs.
+
 **Option 1: Directly Flash to SD Card**
 
 ```bash
--- a/conf/machine/jetson-agx-orin-devkit-emmc-wendyos.conf
+++ b/conf/machine/jetson-agx-orin-devkit-emmc-wendyos.conf
@@ -99,5 +99,13 @@ IMAGE_FSTYPES:pn-tegra-initrd-flash-initramfs:tegra = "${TEGRA_INITRD_FLASH_INIT
 # (P3701-0008, the Industrial 64GB SKU which uses the same eMMC class).
 EMMC_SIZE = "63652757504"
 
+# Root filesystem image size (MB). Overrides the 8192 (8GB) default from
+# recipes-core/images/wendyos-image.bb. flash-image-sizes.inc sizes the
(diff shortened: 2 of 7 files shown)


Joannis:
I stil need to test this properly, let's land this in a nightly if it looks okay - we'll discuss tomorrow

PR #131, opened 2026-06-17
feat(tegra): bring up Bluetooth on Jetson AGX Thor [UNVERIFIED]

Joannis:
> ⚠️ **UNVERIFIED — best-effort, not built or flashed.** Needs validation on AGX Thor hardware before merge. Opened as a draft for that reason.

## Problem

BLE is unavailable on the Jetson AGX Thor (`wendyos-true-raven`, WendyOS 0.15.4). The `wendy-agent` BLE error was the entry point:

```
BLE advertising unavailable error=register advertisement: Method "RegisterAdvertisement"
with signature "oa{sv}" on interface "org.bluez.LEAdvertisingManager1" doesn't exist
```

Diagnosed end-to-end over the agent's own D-Bus connection (via `bluetooth_scan` + the agent's OTLP telemetry — no SSH, no injected logs):

- Powering the adapter fails with `Method "Set" ... on interface "org.freedesktop.DBus.Properties" doesn't exist`. Every D-Bus object implements `Properties`, so this means **`/org/bluez/hci0` does not exist** — BlueZ has **no adapter**.
- `GetManagedObjects` carries **no** `org.bluez.LEAdvertisingManager1` on any path.

So `bluetoothd` runs (the image ships `bluez5`), but the **BT controller is never brought up** → no `hci0` → no adapter on D-Bus.

## Root cause

The Tegra image has no Bluetooth hardware bring-up. Contrast with the RPi machines, which work:

| | RPi (works) | Thor (before this PR) |
|---|---|---|
| `MACHINE_FEATURES bluetooth` | ✅ | ❌ |
| Controller attach | `pi-bluetooth` ships `hciuart.service` | ❌ none |
| BT kernel modules / firmware | present | ❌ none enabled |

## Change

Mirror the RPi bring-up for Tegra:

1. **`bluetooth.cfg`** kernel fragment — enables `CONFIG_BT` + USB/UART HCI transports + Realtek/Broadcom vendor drivers (off in the NVIDIA defconfig). Wired into the jp7 `linux-noble-nvidia-tegra` bbappend alongside the existing `usb-gadget` fragments.
2. **`bluetooth-config`** recipe (`meta-tegra-extensions/recipes-connectivity`) — autoloads `btusb` via `modules-load.d` and `RDEPENDS` the BT kernel modules + controller firmware (`linux-firmware-rtl-bt`). Mirrors the `usb-gadget-modules` recipe pattern.
3. **`packagegroup-wendyos-tegra`** installs `bluetooth-config`.
4. **Thor machine conf** sets `MACHINE_FEATURES bluetooth`.

## ⚠️ What needs verifying on hardware

- **Controller identity.** The vendor driver (`BT_RTL`/`BT_BCM`) and firmware package (`linux-firmware-rtl-bt`) **assume a Realtek combo**, as on the AGX Orin DevKit. If the AGX Thor DevKit uses a different controller, swap the driver + firmware package to match.
- **Attach mechanism.** Assumes USB-enumerated BT (`btusb`). If it's UART/serdev-attached, the `hci_uart` transport is enabled as a fallback but will additionally need a `btattach` service with the correct UART node.
- **Kconfig symbols.** Confirm all symbols land in the 6.8 `.config` (no `do_kernel_configcheck` warnings).
- **Acceptance:** after flashing, `hci0` appears, `busctl introspect org.bluez /org/bluez/hci0` shows `LEAdvertisingManager1`, and `wendy-agent` logs `BLE advertisement registered`.

## Related

Agent-side companion (already correct, independent of this): wendylabsinc/WendyOS#1062 — discovers the LE-advertising adapter instead of assuming `hci0` and turns the cryptic D-Bus error into a clear "no adapter implements LEAdvertisingManager1" message. That PR is what localized this to the OS image.

### DIFF
--- a/conf/machine/jetson-agx-thor-devkit-nvme-wendyos.conf
+++ b/conf/machine/jetson-agx-thor-devkit-nvme-wendyos.conf
@@ -5,6 +5,11 @@
 # L4T 38.4.0. Phase 1 (no Mender, no OTA).
 MACHINEOVERRIDES =. "jetson-agx-thor-devkit:"
 
+# Advertise Bluetooth support so BT-gated packaging behaves like the RPi
+# machines. The hardware bring-up (kernel modules, firmware, hci0 autoload)
+# is pulled in by bluetooth-config via packagegroup-wendyos-tegra.
+MACHINE_FEATURES:append = " bluetooth"
+
 # Inherit from upstream meta-tegra Thor DevKit machine config
 # (P3834-0008 module on P4071-0000 carrier). Pulls in
 # conf/machine/include/tegra264.inc which sets SOC_FAMILY = "tegra264",
--- a/meta-tegra-extensions-jp7/recipes-kernel/linux/linux-noble-nvidia-tegra/bluetooth.cfg
+++ b/meta-tegra-extensions-jp7/recipes-kernel/linux/linux-noble-nvidia-tegra/bluetooth.cfg
@@ -1,0 +1,30 @@
+# Bluetooth stack bring-up for Jetson AGX Thor.
+#
+# UNVERIFIED on Thor hardware. The AGX Thor DevKit's wifi/BT module had not
+# been confirmed at authoring time; these symbols enable the BlueZ-facing
+# kernel side for the controller family used on the AGX Orin DevKit (a Realtek
+# combo whose BT half enumerates over USB), plus the generic UART transport as
+# a fallback for serdev-attached controllers. If the board ships a different
+# controller, adjust the vendor driver here and the firmware package in
+# bluetooth-config to match. Without this fragment the NVIDIA defconfig leaves
+# the BT subsystem off, so no hci0 device is created and BlueZ exposes no
+# adapter (Adapter1 / LEAdvertisingManager1 absent on D-Bus).
+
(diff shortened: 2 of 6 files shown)


conf/machine/jetson-agx-thor-devkit-nvme-wendyos.conf
```diff
@@ -5,6 +5,11 @@
 # L4T 38.4.0. Phase 1 (no Mender, no OTA).
 MACHINEOVERRIDES =. "jetson-agx-thor-devkit:"
 
+# Advertise Bluetooth support so BT-gated packaging behaves like the RPi
+# machines. The hardware bring-up (kernel modules, firmware, hci0 autoload)
+# is pulled in by bluetooth-config via packagegroup-wendyos-tegra.
+MACHINE_FEATURES:append = " bluetooth"
```
abularca, line 11: I've just checked on Thor, `MACHINE_FEATURES` already contains the `bluetooth` feature:
```
dev@yocto:~/wendyos/build$ bitbake wendyos-image -e | grep ^MACHINE_FEATURES=
MACHINE_FEATURES="alsa usbhost pci rtc cuda ext2 wifi bluetooth rtc qemu-usermode"
dev@yocto:~/wendyos/build$
```

PR #150, opened 2026-07-01
Add chained bridge+mesh CNI plugin (wendy-mesh) + Yocto packaging

Joannis:
## What

Adds a thin, stateless **`wendy-mesh` CNI plugin** (Go) chained after the existing `bridge → portmap` conflist, plus its Yocto packaging. It gives a container **per-container mesh egress** to services on other devices, gated by a mesh entitlement — augmenting local bridge networking rather than replacing it.

This is the **in-repo half** of the design. The control plane (peer discovery LAN-first→cloud-broker rendezvous, mTLS data plane over Cloud PKI, service-proxy VIPs, per-node IPAM) lives in `wendylabsinc/WendyOS` and reaches this plugin only through a **versioned CNI `runtimeConfig` contract** — so the plugin shares no code with the agent.

## Design

- **Chained + stateless:** `bridge → portmap → wendy-mesh`; bridge/portmap untouched. The plugin holds no state — every input arrives on stdin.
- **Unmeshed = strict passthrough:** a non-mesh container is a byte-faithful `prevResult` passthrough that never errors and never touches the host.
- **Meshed = fail closed:** adds a `serviceCIDR → gateway` route in the container netns and a host iptables ACCEPT rule scoped to the container's `/32`; any failure returns an error (no silent partial setup, no panic on absent `prevResult`).
- **Forward-compatible:** the service-proxy/VIP model is L4-aware so UDP-over-QUIC can be added agent-side later with **no plugin change**.
- **Opt-in:** packaged only when `WENDYOS_MESH = "1"` (default off); no new kernel modules.

## Testing

TDD throughout. 12 Go tests pass, including the netns-route + iptables tests run for real (`CAP_NET_ADMIN`) in a privileged `linux/arm64` container — `config` (contract parsing/version guard), `mesh` (Add/Del/Check, route + rule idempotency), and `cmd` (passthrough + meshed-absent-prevResult fail-closed). Offline vendored build (`-mod=vendor`) verified; `go.mod` pinned at `go 1.22` with committed `vendor/`.

## ⚠️ Build-host verification still required before merge

`bitbake` was unavailable in the dev environment, so the recipe + packaging (the last two commits) were validated statically (fetch2 layout traced against `go.bbclass`, LICENSE md5, offline build) but **not built**. Please run on a Yocto build host:

- `bitbake -e wendy-mesh-cni | grep -E '^(S|WORKDIR)='` (confirm module lands at `${S}/src/${GO_IMPORT}`)
- `bitbake wendy-mesh-cni` (confirm `do_install` → `/opt/cni/bin/wendy-mesh`)
- `bitbake -e packagegroup-wendyos-container | grep '^RDEPENDS'` off, and `WENDYOS_MESH=1 …` on (gate check)
- `WENDYOS_MESH=1 bitbake wendyos-image`

https://claude.ai/code/session_019BzFgiD7M9XsVAjtWvEPpp


Joannis:
Superseding this in favor of an **agent-only** approach: the mesh route + host iptables ACCEPT rule are applied directly by wendy-agent in its `CNIAdd`/`CNIDel` path (it already has the container IP, netns path, and host iptables access), so a separate chained CNI plugin + Yocto recipe + packaging isn't needed. The `mesh` package logic and its tests port into the agent. Closing; the branch remains for reference.

PR #163, opened 2026-07-02
Pin rootfs to fixed per-platform sizes (8 GiB RPi, 12 GiB Jetson)

Joannis:
## Summary

Rootfs images were content-sized on the wendy-OTA machines, so their size drifted build to build (nightly != release) while the on-disk A/B slots were fixed independently — on Thor at the upstream 25 GiB default the image could silently outgrow, failing mid-write on the device during OTA.

This pins one constant per platform family — **12 GiB Jetson, 8 GiB Raspberry Pi** — as the single source of truth (`conf/distro/include/wendyos-rootfs-size.inc`, required via `wendyos-update.inc`), so the rootfs ext4, the on-disk A/B slot, and the `.wendy` OTA payload are all exactly that size on every build. Nightly and release artifacts are byte-identical in size by construction.

- `IMAGE_ROOTFS_SIZE == IMAGE_ROOTFS_MAXSIZE` (floor == ceiling): exact size while content fits, loud build failure when it doesn't. `EXTRA_SPACE` zeroed (oe-core adds it *after* the floor — the unqualified `:append` in wendyos-image.bb is gated on the pin or it would push every pinned build past the ceiling), overhead factor 1.0.
- **Thor**: `ROOTFSPART_SIZE_DEFAULT = 2 x 12 GiB` so each APP slot exactly matches the image (was the upstream 50 GiB / 25 GiB-per-slot default; the stale "slots are sized from the rootfs image" comment was wrong and is fixed). Reclaimed NVMe space goes to the auto-grown data partition.
- **RPi**: wks rootfsA/B slots move `--size` → `--fixed-size` (overflow fails the build instead of silently growing a slot). RPi3/4 gain the image pin they were missing — they built content-sized images against 8 GiB slots. **Minimum media becomes 32 GB.**
- **Enforcement**: pack-time size assert in `image_types_wendy.bbclass` + a `Verify pinned rootfs size` CI step asserting the deployed ext4 and `.wendy` manifest `payload.size` on every matrix build (PRs, nightlies, releases).

Also folds in a build fix (independent of sizing): `disable_vt_login` masked `getty@.service`, which makes `systemctl --root preset-all` fail `do_rootfs` on every `WENDYOS_DEBUG=0` build — latent since #158 (main stays green because it builds with debug=1; every PR/release image build was broken). Replaced the mask with a disable preset + `ConditionPathExists` drop-in.

**Stacked on top:** #164 removes Mender entirely and pins all Jetson slots (not just Thor) to 12 GiB. Review this one first.

Companion (separate repo): the wendyos-update device-side partition-capacity pre-flight.

Full design + verification checklist: `docs/plans/fixed-rootfs-sizes.md`.

## Test plan

- [x] CI matrix build green (as part of the combined branch, run 28593385417): ext4 == payload == 12884901888 (Jetson) / 8589934592 (RPi) on every device, size step confirmed
- [ ] `wendy install` a Thor: `lsblk -b` shows APP/APP_b == 12884901888
- [ ] RPi5 32 GB SD boots; data grows to fill the card

https://claude.ai/code/session_01LMdFdKRrDpAbT8yLDyWgEm


Joannis:
Second commit: WendyOS is now wendyos-update-only — Mender removed from the active flow, and **all** Jetson A/B slots (Thor + every Orin) are pinned to exactly 12 GiB via `ROOTFSPART_SIZE_DEFAULT = 2 × 12 GiB`. `WENDYOS_OTA` now selects `wendy|none` (default wendy, none for QEMU). Details in the updated `docs/plans/fixed-rootfs-sizes.md`.

PR #191, opened 2026-07-12
feat!: wendyos-update only — remove Mender, pin all Jetson slots to 12 GiB

Joannis:
## What

WendyOS now has a single OTA stack. `WENDYOS_OTA` selects `wendy|none` (default: `wendy` on real hardware, `none` for QEMU); the `mender` option and everything behind it is removed.

- Delete `mender.inc`, `flash-image-sizes.inc`, `mender-secrets.inc`, the mender bblayers includes, `recipes-mender/` (mender-esp, JP6 state scripts), and `mender-config-before-data.bbclass`
- Orin machine confs (AGX nvme/emmc, Nano sd/nvme) drop the dual-stack `IMAGE_FSTYPES`/`IMAGE_NAME_SUFFIX` conditionals and all `MENDER_*` vars, and pin `ROOTFSPART_SIZE_DEFAULT = 2 × 12 GiB` so every Jetson A/B slot matches the pinned rootfs image
- RPi common include + machine confs are wendy-native; the mender-only fstab fixup and sdimg-reorder wiring are gone
- `tegra_partition_config.bbclass` loses the mender data-partition branch and `MENDER_DATA_PART_NUMBER` fallback
- `bootstrap.sh` / CI prefetch no longer clone `meta-mender{,-community}`
- `bootloader-update` moves from `recipes-mender/` to `recipes-bsp/` (it is not Mender-specific — it stages the UEFI capsule for the wendyos-update tegrauefi connector)

Mender references **outside** the OS image config are purged too: the publisher no longer recognises `.mender` artifacts, `system-status.sh` drops its Mender-era slot diagnostics, the CI OTA glob matches only `.wendy`, and stale Mender comments across the board/config files are reworded to describe wendyos-update.

## Notes

- Rebased onto current `main` (0.17.0). The fixed-rootfs-size (#166) and VT-login work landed independently on main, so those two commits were dropped — only the Mender removal remains.
- Net: **110 files, +322 / −2079**.
- `go test ./tools/publisher/...` passes; shell scripts pass `sh -n`.
- **Breaking:** the last Mender-capable tree is the pre-blacksail `main` history.


Joannis:
Previously approved by Adrian and Andrei - but merged into a branch that is now closed

PR #199, opened 2026-07-16
flash-init: parse device identity from suffixed DTB compatibles (WDY-1888)

EBro912:
## Problem

Orin Nano recovery flashes fail at the host's identity gate:

```
✗ wrong Jetson hardware: detected module PUNKNOWN- on carrier PUNKNOWN-; flashpack requires module P3767-0005 on carrier P3768-0000
```

The recovery initrd's `write_device_identity` (added in #189) parses `/proc/device-tree/compatible` with a regex anchored to exactly `nvidia,p<carrier>-<sku>+p<module>-<sku>`. The Orin Nano (Super) devkit DTB's root compatible is `nvidia,p3768-0000+p3767-0005-super` — the `-super` suffix breaks the `$` anchor, the parse falls through to `UNKNOWN/UNKNOWN`, and the Wendy host (correctly) refuses every Orin Nano flash. AGX Orin was unaffected because its compatible (`nvidia,p3737-0000+p3701-0005`) carries no suffix.

## Fix

Allow a DTB variant suffix on either half of the compatible and extract only the four numeric fields. An unparsable compatible still degrades to `UNKNOWN`, which the host still refuses.

## Verification

- Parse logic exercised against the live Nano flashpack DTB compatible (`nvidia,p3768-0000+p3767-0005-super` → `3768-0000` / `3767-0005`), the AGX compatible (unchanged), a hypothetical carrier suffix, and garbage input (→ `UNKNOWN`).
- Patch re-applies cleanly against the pinned meta-tegra `init-flash.sh` (`git apply --check`, hunks land with small offsets) and the patched script passes `sh -n`.
- On-device verification pending: needs this PR's flashpack artifact (`wendy os install --device-type jetson-orin-nano --pr <this PR>`); found live-debugging an Orin Nano 8GB devkit on macOS.

### DIFF
--- a/meta-tegra-extensions/recipes-core/initrdscripts/tegra-flash-init/0001-flash-init-export-device-identity.patch
+++ b/meta-tegra-extensions/recipes-core/initrdscripts/tegra-flash-init/0001-flash-init-export-device-identity.patch
@@ -7,10 +7,15 @@ Expose the exact T234 module/carrier tuple and USB session in the first
 flashpkg LUN. The Wendy host validates this before replacing the LUN, which is
 the destructive QSPI/rootfs handoff boundary.
 
+The compatible parse tolerates DTB variant suffixes: the Orin Nano Super
+devkit's root compatible is "nvidia,p3768-0000+p3767-0005-super", which an
+exact-anchored match reported as UNKNOWN/UNKNOWN, making the host refuse
+every Orin Nano flash.
+
 Upstream-Status: Inappropriate [wendy-specific]
 ---
- init-flash.sh | 30 +++++++++++++++++++++++++++++-
- 1 file changed, 29 insertions(+), 1 deletion(-)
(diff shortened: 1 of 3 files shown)


EBro912:
On-device verified: a full `wendy os install --device-type jetson-orin-nano --pr 199` recovery flash on an Orin Nano 8GB devkit (macOS host) passed the identity gate — `identity verified: module P3767-0005, carrier P3768-0000` — and completed QSPI + NVMe with final status SUCCESS. The board boots and is discoverable.

One unrelated first-boot issue surfaced during verification (device came up with mDNS `WENDY_DISPLAY_NAME` placeholders because `wendyos-uuid-generate`/`wendyos-device-name-generate` never got start jobs on the very first boot; a reboot healed it) — being investigated separately, not caused by this PR.

PR #209, opened 2026-07-29
wendyos-agent: rotate binary backups + fix nightly no-op reinstall (WDY-2185)

chrisdok43:
Fixes [WDY-2185](https://linear.app/wendylabsinc/issue/WDY-2185/nightly-agent-updater-leaks-unpruned-binary-backups-until-the-root-fs): the nightly agent auto-updater fills the root filesystem until `wendy cloud device update` (and the OS-update check behind it) fails with ENOSPC. Found on wendy-box-theta (ENMAX pilot): 68 × 22MB `wendy-agent.backup.*` files = 1.3GB, root fs at 100% with 16MB free.

## Fix 1 — backup rotation

`download-wendyos-agent.sh` kept a timestamped 22MB backup on **every install, forever**. Now it prunes to the newest 2 after each backup (`KEEP_BACKUPS=2`) — the restore path only ever reads `${BINARY_NAME}.latest`, so timestamped copies are purely forensic.

`wendyos-agent-updater.sh` additionally prunes on **every nightly check** (even "already up to date"), so already-deployed devices with a backlog self-heal on the first timer run after an OS image update ships this script — no manual fleet cleanup.

## Fix 2 — why it leaked *nightly* rather than per-release

`get_current_version()` truncated the calver version to its date part (`2026.07.27-003050` → `2026.07.27`, regex only matched `x.y.z`) while release tags keep the time suffix. Current never equalled latest, so `version_gt` reported an update **every night** and the updater re-downloaded + re-installed the *same* release — each run adding a backup. The regex now captures the full version including the `-HHMMSS` suffix; plain semver matches unchanged, and dev builds still exit early on the raw-string check (WDY-1770 behavior preserved).

## Verification

- `bash -n` on both scripts
- Prune fixture test: 68 fake backups → exactly 2 remain; empty/missing backup dir is a safe no-op under `set -e` (pipeline ends in `while`, glob failure suppressed)
- Regex: `2026.07.27-003050` captured in full; `1.0.3` unchanged; `2026.06.30-133859-dev` → numeric part only (dev check uses the raw string and runs before the version comparison)
- Equal full versions now compare "already up to date" → no reinstall, no backup

## Not in this PR (noted in WDY-2185)

- Immediate remediation of already-full devices (theta needs a one-time manual prune or this script arriving via OS image update — the A/B slot rewrite also clears the backlog)
- Moving `BACKUP_DIR` off the root fs to `/data`
- Root-fs image built at 7.3GB inside its 14GB A/B slot

### DIFF
--- a/recipes-core/wendyos-agent/files/download-wendyos-agent.sh
+++ b/recipes-core/wendyos-agent/files/download-wendyos-agent.sh
@@ -208,6 +208,11 @@ install_binary() {
 
     log "Created required directories"
 
+    # Free space before the ~22MB copy below, which on a full root fs fails
+    # under `set -e` before any prune runs. Outside the -f guard so a device
+    # with no installed binary still sheds a backlog.
+    prune_backups
+
     # Backup existing binary if it exists
     if [ -f "${INSTALL_DIR}/${BINARY_NAME}" ]; then
         log "Backing up existing binary"
--- a/recipes-core/wendyos-agent/files/download-wendyos-agent.sh
+++ b/recipes-core/wendyos-agent/files/download-wendyos-agent.sh
@@ -156,6 +156,22 @@ download_binary() {
     log "Binary downloaded and prepared successfully"
 }
 
+# Prune old timestamped backups, keeping the newest N. Without this the
+# nightly updater leaks one ~22MB copy per run until the root filesystem
+# fills and agent/OS updates start failing with ENOSPC (WDY-2185). The
+# restore path only ever uses ${BINARY_NAME}.latest; timestamped copies are
+# forensic convenience, so a couple is plenty.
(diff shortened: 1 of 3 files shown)


Joannis (approved):

thombles:
I fixed the merge conflict and made some tweaks:
* #208 had a fuller fix for the version-parsing with suffix, including -dev builds and unit tests so I've taken that one in preference. Without the `-dev` tracking we would clobber dev builds by accident again.
* To pass the new shellcheck CI, changed the prune function to use globbing on the filename timestamps, which is in a predictable order
* Added a prune before trying to back up the existing binary to help recover a machine that's already full
* Added pruning to the test suite

PR #215, opened 2026-08-05
publisher: scope upload timeouts per file instead of one shared 30m context

EBro912:
## Why

The 0.18.1 release ([run 30573040970](https://github.com/wendylabsinc/WendyOS-Builder/actions/runs/30573040970), attempt 1) shipped without `jetson-orin-nano`: the publisher wraps **all** of a device's uploads in a single 30-minute context, and on a degraded runner link the Orin Nano's ~17.6 GB of artifacts drained it. The recovery bundle's last part finalize died with `context deadline exceeded`, the job failed after the image and OTA had already uploaded, and the device's manifest stayed on 0.17.0 while every other device advanced to 0.18.1 (fixed operationally by re-running the job).

This was not a one-device anomaly. Upload-step durations (minutes) across recent runs:

| Device | 4 nightlies | 0.18.1 attempt 1 | 0.18.1 attempt 2 |
|---|---|---|---|
| jetson-orin-nano (nvme sd) | 5.8 – 13.9 | **59.8 (failed at deadline)** | 9.7 |
| jetson-agx-orin (nvme emmc) | 5.2 – 9.9 | **28.8** | — |
| jetson-agx-thor (nvme) | 2.2 – 9.1 | 5.2 | — |
| generic-x86-64 (disk) | 1.2 – 1.8 | 6.6 | — |
| raspberry-pi-3/4/5 | 0.6 – 1.7 | 1.0 – 1.4 | — |

`jetson-agx-orin` came within ~1 minute of the same wall in the same run (its emmc-variant invocation spent ~19.6 of its 30 minutes on uploads alone). The composite uploader was designed around ~113 MB/s; that evening the runners saw ~4–7 MB/s per file. Same payload, next attempt: 9.7 minutes — the variance is network, not payload, so a fixed shared budget is the wrong shape.

## What

- Each file upload now gets its own deadline scaled to its size: transfer time at a 2 MiB/s floor throughput + 5 min headroom, never below 10 min. The 6.9 GB recovery bundle that died gets ~60 min to itself; small files keep tight bounds.
- The whole-run context becomes a coarse 2-hour safety net (it must exceed the largest single-file budget, since a child context cannot outlive its parent; still well under the 6 h job timeout).
- Deadline failures now name the file and the budget it exceeded instead of a bare `context deadline exceeded`.

## Testing

- `go test ./...` in `tools/publisher` (new `TestUploadTimeoutFor` covers the floor, the scaling formula, the incident's exact file size, and monotonicity).
- `go vet`, `gofmt` clean.

### DIFF
--- a/tools/publisher/parallel_upload.go
+++ b/tools/publisher/parallel_upload.go
@@ -9,6 +9,7 @@ import (
 	"path/filepath"
 	"strings"
 	"sync"
+	"time"
 
 	"cloud.google.com/go/storage"
 	"github.com/sirupsen/logrus"
@@ -40,6 +41,16 @@ const (
 	// flushes; 32 MiB keeps the socket busier. This is a per-part in-flight
 	// buffer, not the whole part, so memory stays bounded.
 	partChunkSize = 32 << 20 // 32 MiB
+
+	// Per-file upload timeout parameters for uploadTimeoutFor. The floor
+	// throughput is deliberately far below the ~113 MB/s a healthy runner
+	// link sustains: it is the worst degradation we tolerate before giving
+	// up, sized so that concurrent big files sharing a saturated link still
+	// fit their individual budgets (the 0.18.1 release saw ~4-7 MB/s per
+	// file and blew the old shared 30-minute context).
+	uploadTimeoutFloor     = 10 * time.Minute
+	uploadTimeoutHeadroom  = 5 * time.Minute
+	uploadFloorBytesPerSec = 2 << 20 // 2 MiB/s
 )
 
 // crc32cTable is the Castagnoli table GCS uses for its CRC32C object checksums.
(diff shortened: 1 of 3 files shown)


thombles (commented):
**Issue 1 - Changing timeout from 30 to 120 mins exposes token-handling bugs**

This refresh system has never been exercised for real because of the 30 minute limit.

- A gcloud token is generated at the start of the upload step. It has validity of 60 minutes.
- The same token is handed to different invocations of the publisher binary in a loop for different storage formats.
- Each instance assumes (hard-coded) that it will be valid for at least 55 minutes, which it may not be
- When it decides a refresh is required it calls print-access-token but this won't necessarily work if the internal cache determines that the token is fresh enough (>3m45s remain)

**Solution** - stacked PR https://github.com/wendylabsinc/WendyOS-Builder/pull/229:
- Use the authentication and token refresh built in to the go google cloud client and stop trying to handle and refresh tokens manually
- Important: Because this publisher is checked out and used in the build process of wendy-lite, it also must be migrated. We need to merge PR https://github.com/wendylabsinc/wendy-lite/pull/33 first.

**Issue 2 - per-file timeouts aren't timed correctly**

There is a semaphore gating the number of concurrent uploads, which includes multiple parts of the same file being uploaded concurrently. This means that even small files might take a while to get a chance to upload. The minimum 10 minutes starts counting before that semaphore is acquired. There is evidence in the logs of tiny files waiting over 5 minutes, so it's possible they will sometimes time out without ever getting to start.

The commit message says the goal is "one slow multi-GB upload can't starve the others" but this doesn't really do that since they're split and run in parallel. A single large file's 16 parts can fill the semaphore on their own. With three large files wanting 48 slots and 16 available, each file effectively gets a third of the streams, so its wall time is roughly 3x a solo upload while its budget was computed as though it were the only thing running.

**Suggestion:** The per-file limits are more trouble than they're worth IMO. Just putting a 2h cap on everything will do the job.

PR #218, opened 2026-08-06
guarantee native gamepad kernel support

EBro912:
## Summary

- add one shared game-controller Kconfig contract to Jetson JP6/JP7, Raspberry Pi, and shipping x86 kernel recipes
- guarantee evdev/joydev, generic and USB HID, UHID, Bluetooth HIDP, xpad, and Microsoft HID support as built-ins or shipped modules
- recommend modular driver packages in the image while excluding QEMU from the hardware contract
- validate each effective kernel config and finished image manifest in build CI
- document the BlueZ-owned Bluetooth/USB smoke-test workflow

## Why

Native controller access through the existing `input` entitlement still depends on each BSP producing the required Linux input and HID interfaces. A config fragment by itself is not sufficient because vendor Kconfig dependencies can silently drop a requested symbol.

The shared include checks the post-Kconfig result, deploys that exact config beside the image, and CI confirms every `=m` driver package is actually present in the finished image.

## Impact

Shipping Jetson, Raspberry Pi, and x86 images expose the upstream Linux gamepad interfaces needed by entitled services. BlueZ remains responsible for pairing, trust, and reconnect; this adds no controller daemon, mappings, profiles, or rumble API.

This is the second PR in the rollout order, after WendyOS Agent/CLI PR #1589 and before the Rosmaster sample update.

## Validation

- `PYTHONDONTWRITEBYTECODE=1 python3 -m unittest scripts/check_game_controller_support_test.py` (7 tests)
- parsed both modified workflow YAML files
- `git diff --check`

A full Yocto kernel/image build and the Bluetooth/USB smoke test require CI or target hardware and remain acceptance checks for this draft.

PR #219, opened 2026-08-07
Swap Orin Nano Bluetooth to mainline btusb+btrtl

EBro912:
## Summary

- add an opt-in per-machine Bluetooth kernel contract (`wendy-bluetooth.inc`, same shape as the game-controller contract from #218) with required and forbidden symbol validation and a deployed evidence config
- enable it for both Jetson Orin Nano devkit machines with a btusb+btrtl transport fragment
- remove the NVIDIA out-of-tree `rtk_btusb` driver at the package level (machine-conf overrides + `BAD_RECOMMENDATIONS`) and guarantee it can never bind first via a modprobe blacklist recipe
- ship `linux-firmware-rtl8822` (the `rtl_bt/` layout mainline btrtl loads) through tegra-packagegroup-base, and recommend the =m Bluetooth module packages through packagegroup-wendyos-kernel
- validate in CI with `scripts/check_bluetooth_stack.py`: a BlueZ ≥5.80 floor on every image (the blacksail pin ships 5.86), and on contract machines the effective kernel config, required packages, and the absence of the vendor packages

## Why

The devkit's RTL8822CE radio (USB 0bda:c822) drops BLE HID links in a metronomic ~2 s connect/drop loop under `rtk_btusb`: observed on the Rosmaster car as 62 uhid create/destroy cycles in 122 s, with one kernel "Bad flag" line per cycle and the ring buffer wrapped by `rtk_btcoex` debug spam. The same hardware+driver combo is broken on stock JetPack with no known fix (NVIDIA forum 274090). Wired USB proves everything above the transport is healthy, and the image already runs BlueZ 5.86 — the vendor transport is the remaining lever.

AGX Orin and Thor intentionally keep the vendor driver until each gets its own validated swap: enabling btusb while rtk_btusb is installed would race probes on the same USB ID.

## Dependencies and rollout

Stacked on #218 (`agent/native-gamepad-kernel`) — merges after it. The companion agent-side pairing-flow fix is wendylabsinc/WendyOS#1595.

## Validation

- `python3 -m unittest scripts/check_bluetooth_stack_test.py` — 18 tests
- `python3 -m unittest scripts/check_game_controller_support_test.py` still green
- both workflow files parse
- `git diff --check`

## Hardware acceptance still required

- flash the Orin Nano car; dmesg shows btusb probing 0bda:c822 and `RTL: firmware rtl_bt/rtl8822cu_fw.bin`, zero rtk_* lines
- `wendy device bluetooth` scan/connect/forget; BLE discovery advert + L2CAP mTLS channel; A2DP via wireplumber; a bluetooth-entitled app through the D-Bus proxy
- Xbox controller fresh pair, evdev arrival, 30-minute soak with concurrent wlan0 iperf3 (coex watch), controller power-cycle auto-reconnect
- warm reboots and an A/B OTA from a vendor-driver build

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1036,6 +1036,30 @@ jobs:
             python3 scripts/check_game_controller_support.py --config "$CONFIG" --manifest "$MANIFEST"
           done
 
+      # Same evidence pattern for the Bluetooth stack: every image must meet
+      # the BlueZ floor, and machines under the mainline-transport contract
+      # (wendy-bluetooth.inc) must additionally show btusb+btrtl in the
+      # effective config, the rtl_bt firmware + modprobe blacklist packages
+      # present, and the vendor rtk_btusb packages absent. The evidence
+      # config exists only where the contract is enabled; the validator
+      # itself fails contract machines that are missing it.
+      - name: Validate Bluetooth stack
+        working-directory: ${{ github.workspace }}/wendyos
+        env:
+          MAP: ${{ steps.vars.outputs.map }}
+        run: |
+          set -euo pipefail
+          for token in $MAP; do
+            IFS='|' read -r storage board machine <<< "$token"
+            DEPLOY_DIR="$GITHUB_WORKSPACE/build/tmp/deploy/images/$machine"
+            CONFIG="$DEPLOY_DIR/wendyos-bluetooth-$machine.config"
+            MANIFEST=$(find "$DEPLOY_DIR" -maxdepth 1 -name 'wendyos-image-*.manifest' ! -type l -print -quit)
+            [[ -n "$MANIFEST" ]] || { echo "::error::missing image manifest in $DEPLOY_DIR"; exit 1; }
+            ARGS=(--machine "$machine" --manifest "$MANIFEST")
+            [[ -f "$CONFIG" ]] && ARGS+=(--config "$CONFIG")
+            python3 scripts/check_bluetooth_stack.py "${ARGS[@]}"
+          done
+
       # The persistent NVMe cache is written in place by the build (sstate,
       # downloads, repos are symlinked onto it) — no upload/save step needed.
 
--- a/.github/workflows/scripts-ci.yml
+++ b/.github/workflows/scripts-ci.yml
@@ -11,6 +11,7 @@ on:
       - 'scripts/make-jetson-disk-img.py'
       - 'scripts/make_jetson_disk_img_test.py'
       - 'scripts/check_game_controller_support*'
+      - 'scripts/check_bluetooth_stack*'
       - 'recipes-core/wendyos-agent/files/*.sh'
       - '.github/device-artifacts.json'
       - '.github/workflows/scripts-ci.yml'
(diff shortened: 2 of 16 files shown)


EBro912:
Hardware validation on the devkit (flashed 0.18.1 PR image, Orin Nano NVMe, kernel 6.8.12-l4t-r39.2.0):

**What works**
- `rtk_btusb`/`rtk_btcoex` are gone; `btusb`+`btrtl`+`btintel`/`btmtk`/`btbcm` load as modules; BlueZ 5.86; `rtl_bt/rtl8822cu_fw.bin` + `_config.bin` shipped. The PR 218 input contract also holds on this image (`/dev/uhid` present, `joydev`/`hid_microsoft` builtin).

**What didn't: no hci0 at all** → fixed by c70aea5
- The radio (USB `0bda:c822`) enumerates with textbook BT endpoints, but `/sys/class/bluetooth` stays empty and a manual sysfs bind returns "No such device". Cause: NVIDIA's kernel marks `0bda:c822` `BTUSB_IGNORE` in **both** btusb device tables (read out of the shipped `btusb.ko`: both entries `driver_info=0x1`, siblings `0x210000`) so their vendor driver can claim it. With the vendor driver removed, nothing can bind the radio. c70aea5 adds the machine-gated kernel source patch restoring the mainline `BTUSB_REALTEK | BTUSB_WIDEBAND_SPEECH` entry; verified against the pinned kernel source (`linux-noble` @ 7827294) with strict `git apply`. Needs the rebuilt image to hardware-verify.

**New regression to track: Wi-Fi is dead on this image**
- `wlan0` never appears. The PCIe half of the RTL8822CE (`0001:01:00.0`, `10ec:c822`) is detected and the vendor `rtl8822ce` wifi module loads, but probe fails with `power init fail` (visible on modprobe cycle). Suspected interplay with this PR's vendor-BT removal (shared power/coex on the combo chip — possibly just the BT function being unclaimed at wifi power-on; the pre-PR image with `rtk_btusb` had working wifi). Worth retesting once the btusb fix gives the BT function a driver again; if wifi still fails, the likely completion of this swap is mainline `rtw88_8822ce` for the wifi half too. Until then the devkit is USB-gadget-only for networking.

EBro912:
Follow-up from tonight's hardware session, one honesty note for this PR's premise: with the c70aea5 fix applied on-device (btusb binding, RTL fw loaded, BlueZ 5.86), we reproduced a **~2 s connect/drop loop with a fresh uhid node per cycle on the mainline stack** — byte-for-byte the symptom this PR attributes to `rtk_btusb`. Cause per btmon: the Xbox controller (fw 5.09) discards its own bond minutes after pairing, so every bluetoothd auto-reconnect dies at `LE Start Encryption → PIN or Key Missing (0x06)` → local disconnect → retry ~2 s later.

That means the original vendor-driver loop may have been (partly or wholly) the same controller-firmware pathology, observed without HCI-level capture at the time. The swap remains justified on its own merits — mainline maintainability, the BlueZ floor contract, no rtk_btcoex dmesg spam, and tonight it demonstrably pairs/encrypts/attaches HID correctly — but "fixes the 2 s loop" should not be claimed until re-tested with updated controller firmware. Will re-validate and report here.

Also still open from the earlier comment: the Wi-Fi `power init fail` regression (wlan0 absent).

EBro912:
Closing, the original issues allegedly fixed by this PR were controller firmware issues, the swap to btrtl is not necessary. Still worth considering in the future as mainline btusb + btrtl drives are the modern standard, however this would also require similar WiFi changes as the stock WiFi driver dies without its companion stock BT driver. PR #218 is still necessary to solidify gamepad input support on the Nano.

PR #224, opened 2026-08-10
fix(identity): let an explicit rename outrank device-name in the mDNS records

Joannis:
## The bug

`wendy device rename` does not stick across a reboot — the device keeps resolving as `<new>.local`, but `wendy device list` shows the pre-rename name again.

A rename writes the chosen name to two places:

| What | Persists? | Why |
|---|---|---|
| Hostname (`<name>.local`) | ✅ | `/etc/wendy-agent/hostname` is bind-mounted to `/data`, and `generate-hostname.sh` prefers it verbatim on boot. |
| mDNS TXT `name` / `displayname` | ❌ | `update-mdns-uuid.sh` re-derives them from `/etc/wendyos/device-name` on **every** boot. |

`wendy device rename` never writes `device-name` — it persists to `/etc/wendy-agent/hostname`. So this publisher reverted the advertisement on the next boot. Those two records are what the CLI displays: `lanDeviceFromService` lets the `displayname` TXT record win over the hostname.

## The fix

Prefer `/etc/wendy-agent/hostname` when it holds a valid DNS label, exactly as `generate-hostname.sh` already does for the hostname itself; fall back to `device-name` otherwise. Validation mirrors `is_valid_hostname` there and `validHostname` in the agent, so a corrupt file falls back rather than burning junk into the advertisement.

`device-name` is untouched — it remains the generated factory identity. This only changes which name is *advertised*.

## Tests

Paths are now overridable so the script can run outside a device. `update-mdns-uuid.test.sh` (plain bash, same shape as `wendyos-agent-updater.test.sh`) covers the precedence, the fallbacks, and that records the agent owns at runtime (`tls`, port) are left alone.

Verified the test discriminates: against the old name-selection logic with the same parameterized paths, exactly the 4 explicit-rename assertions fail and the other 12 pass — so it catches this bug specifically, without masking a regression elsewhere.

```
passed: 16, failed: 0
```

It needs GNU sed (the script uses `sed -i -E`) and says so loudly rather than passing silently on macOS; it uses `gsed` when installed.

## Companion PR

wendylabsinc/WendyOS#1653 re-asserts the same records at agent startup. That covers the A/B OTA case — the avahi service file is per-slot rootfs state — and reaches already-deployed devices without an OS update.

## Not verified

No hardware reboot test. The reasoning is from the boot ordering (`wendyos-identity.service` is `WantedBy=multi-user.target`, so it runs every boot) and the persistence map in `setup-etc-binds.sh`.

### DIFF
--- a/recipes-core/wendyos-identity/files/update-mdns-uuid.sh
+++ b/recipes-core/wendyos-identity/files/update-mdns-uuid.sh
@@ -1,29 +1,30 @@
 #!/bin/bash
-# WendyOS mDNS identity publisher.
-#
-# Sets the id/name/displayname TXT records in the Avahi service file to the
-# current device identity by replacing each record's VALUE in place — NOT by
-# one-shot placeholder substitution, and NOT by rewriting the whole file.
-#
-# Why value-replacement:
-#   - Idempotent + self-correcting: a value left by an earlier run (the original
-#     WENDY_DEVICE_NAME placeholder, or a stale/`unknown-device` name) is matched
-#     and replaced, instead of being permanently stuck once the placeholder was
-#     consumed. That stuck-placeholder bug is what left mDNS advertising
-#     name=unknown-device on a clean first boot.
-#   - Preserves records the wendy-agent manages at RUNTIME — the service port,
-#     the tls=true/false record (UpdateAvahiForProvisioning), and fqdn
-#     (updateAvahiDeviceName). A from-scratch rewrite here would revert the
-#     provisioned advertisement on the next reboot.
+# WendyOS mDNS identity publisher. Replaces each TXT record's VALUE in place, so
+# a placeholder or stale value from an earlier run is still overwritten and the
+# records the agent owns at runtime (port, tls) survive untouched.
 set -u
 
-UUID_FILE="/etc/wendyos/device-uuid"
-DEVICE_NAME_FILE="/etc/wendyos/device-name"
-SERVICE_FILE="/etc/avahi/services/wendyos-mdns.service"
+# Overridable so the test can run against a temp tree; the unit passes no
+# environment on a device, so these take the defaults.
+UUID_FILE="${UUID_FILE:-/etc/wendyos/device-uuid}"
+DEVICE_NAME_FILE="${DEVICE_NAME_FILE:-/etc/wendyos/device-name}"
+# Literal hostname from 'wendy device rename'. Outranks the generated
+# device-name (as in generate-hostname.sh) so a rename survives a reboot.
+EXPLICIT_HOSTNAME_FILE="${EXPLICIT_HOSTNAME_FILE:-/etc/wendy-agent/hostname}"
+SERVICE_FILE="${SERVICE_FILE:-/etc/avahi/services/wendyos-mdns.service}"
+GENERATE_DEVICE_NAME_SH="${GENERATE_DEVICE_NAME_SH:-/usr/bin/generate-device-name.sh}"
+GENERATE_HOSTNAME_SH="${GENERATE_HOSTNAME_SH:-/usr/sbin/generate-hostname.sh}"
+AVAHI_RELOAD_CMD="${AVAHI_RELOAD_CMD:-systemctl}"
+IDENTITY_WAIT_SECS="${IDENTITY_WAIT_SECS:-10}"
+
+# Shared identity helpers, one source of truth with generate-hostname.sh.
+IDENTITY_LIB="${IDENTITY_LIB:-/usr/share/wendyos/identity-lib.sh}"
+# shellcheck source=/dev/null
+. "$IDENTITY_LIB" || { echo "Cannot source identity helpers: $IDENTITY_LIB" >&2; exit 1; }
 
 # Wait briefly for the identity files (the generators are ordered before us, but
 # stay defensive against a slow /data bind).
-for i in {1..10}; do
+for _ in $(seq "$IDENTITY_WAIT_SECS"); do
     [ -f "$UUID_FILE" ] && [ -f "$DEVICE_NAME_FILE" ] && break
     sleep 1
 done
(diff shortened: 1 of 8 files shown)


recipes-core/wendyos-identity/files/update-mdns-uuid.test.sh
```diff
@@ -0,0 +1,139 @@
+#!/usr/bin/env bash
+# Tests for update-mdns-uuid.sh — plain bash, no framework, same shape as
```
kelenyibenjamin, line 2: the new update-mdns-uuid.test.sh is not wired into any workflow, so the PR's entire test claim is unenforced. Have to be wired in: workflows/scripts-ci.yml

recipes-core/wendyos-identity/files/update-mdns-uuid.test.sh
```diff
@@ -0,0 +1,139 @@
...
+    <txt-record>tls=true</txt-record>
+  </service>
+</service-group>
+EOF
+
+    UUID_FILE="${WORK}/wendyos/device-uuid" \
+    DEVICE_NAME_FILE="${WORK}/wendyos/device-name" \
+    EXPLICIT_HOSTNAME_FILE="${WORK}/wendy-agent/hostname" \
+    SERVICE_FILE="${WORK}/avahi/wendyos-mdns.service" \
+        bash "${SCRIPT}" >/dev/null 2>&1
```
kelenyibenjamin: most of the test uses temporary files, which is good. However, 2 important paths are still hard-coded:

* the generator's real location
* the Avahi reload command

because of this, parts of the test can affect the actual environment where the test is running instead of staying completely inside the temporary test directory

the test should be completely isolated and should not modify the real /etc, Avahi configuration, or running services

recipes-core/wendyos-identity/files/update-mdns-uuid.test.sh
```diff
@@ -0,0 +1,139 @@
...
+    if [[ "${got}" == "${want}" ]]; then
+        ok "${label}: ${key}=${want}"
+    else
+        bad "${label}: ${key}=${got} (want ${want})"
+    fi
+}
+
+# --- (a) an explicit rename wins and is advertised verbatim ------------------
+# The literal hostname is used as-is: no "wendyos-" prefix is derived from it.
+run_publisher "brave-dolphin" "kitchen-pi"
```
kelenyibenjamin, line 143: the actual problem this PR is trying to solve happens when a device has already been renamed and a real old hostname is still present from a previous boot

the current test therefore doesn't verify the most important scenario: device was previously renamed - device-name changes - explicit rename should still win

add tests where:

* arealistic hostname already exists
* the device-name changes
* the explicit hostname should remain the one used by mDNS
* running the update script again does not change anything unnecessarily

recipes-core/wendyos-identity/files/update-mdns-uuid.test.sh
```diff
@@ -0,0 +1,139 @@
...
+assert_txt name        "brave-dolphin" "no rename"
+assert_txt displayname "Brave Dolphin" "no rename"
+
+# --- (c) an invalid explicit hostname falls back to device-name --------------
+# Uppercase is lowercased before validation, so it is accepted, not rejected.
+run_publisher "brave-dolphin" "Kitchen-Pi"
+assert_txt name        "kitchen-pi" "uppercase explicit is lowercased"
+assert_txt displayname "Kitchen Pi" "uppercase explicit is lowercased"
+
+for junk in "-leading-hyphen" "trailing-hyphen-" "1starts-with-digit" "has_underscore" "has.dot"; do
```
kelenyibenjamin: WendyOS uses A/B slots, and after switching to a fresh slot, there is a realistic situation where some identity files contain values from the previous boot while others contain new/default values

the test matrix does not cover this combination

this is important because this scenario also exercises the script's waiting logic and generator recovery behavior

add a test case representing a fresh A/B slot after a rename, where the identity files contain the realistic combination of old/new values

recipes-core/wendyos-identity/files/update-mdns-uuid.sh
```diff
@@ -69,20 +86,36 @@ if [ ! -s "$DEVICE_NAME_FILE" ] && [ -x /usr/bin/generate-device-name.sh ]; then
     fi
 fi
 
-# Only set name/displayname when we actually have a device name; otherwise leave
+# Pick the advertised name: an explicit 'wendy device rename' hostname wins and
+# is used verbatim; otherwise fall back to the generated device name.
+ADVERTISED_NAME=""
+NAME_SOURCE=""
+if [ -s "$EXPLICIT_HOSTNAME_FILE" ]; then
```
kelenyibenjamin, line 82: once /etc/wendy-agent/hostname is created, nothing removes it

that means if a user explicitly renames a device and later changes the device name, the device may continue using the old explicit hostname forever

the runtime behavior may look correct, but after reboot the old hostname can come back because the explicit hostname file still exists

recipes-core/wendyos-identity/files/update-mdns-uuid.sh
```diff
@@ -69,20 +86,36 @@ if [ ! -s "$DEVICE_NAME_FILE" ] && [ -x /usr/bin/generate-device-name.sh ]; then
...
+# Only set name/displayname when we actually have a name; otherwise leave
 # whatever is there (placeholder or a prior good value) for a later run — never
 # burn a junk fallback into the advertisement.
-DEVICE_NAME=""
-if [ -s "$DEVICE_NAME_FILE" ]; then
-    DEVICE_NAME=$(tr -d '[:space:]' < "$DEVICE_NAME_FILE")
-    DISPLAY_NAME=$(echo "$DEVICE_NAME" | sed 's/-/ /g' | awk '{for(i=1;i<=NF;i++) $i=toupper(substr($i,1,1)) tolower(substr($i,2));}1')
-    set_txt name "$DEVICE_NAME"
+if [ -n "$ADVERTISED_NAME" ]; then
+    DISPLAY_NAME=$(echo "$ADVERTISED_NAME" | sed 's/-/ /g' | awk '{for(i=1;i<=NF;i++) $i=toupper(substr($i,1,1)) tolower(substr($i,2));}1')
```
kelenyibenjamin: displayname diverges from the agent, causing one avahi restart per boot

solution: create a helper, like:
```
+display_name() {
+    local -a words
+    IFS='-' read -r -a words <<< "$1"
+    local i
+    for i in "${!words[@]}"; do words[i]="${words[i]^}"; done
+    local IFS=' '
+    printf '%s' "${words[*]}"
+}
```

then:

`DISPLAY_NAME=$(display_name "$ADVERTISED_NAME")`

kelenyibenjamin (commented):
1. the new test never runs in CI
2. make the test fully isolated
3. fix hostname validation/normalization consistency between the script and the agent
4. test the real rename scenario including an already-renamed device and idempotency

feel free to let me know if you’d like me to take care of all the fixes from the review. If you’re okay with it, I can also commit the changes directly to your branch to make things easier

cc: @Joannis

recipes-connectivity/avahi/files/wendyos-mdns.service
```diff
@@ -8,5 +8,6 @@
     <txt-record>id=WENDY_DEVICE_ID</txt-record>
     <txt-record>name=WENDY_DEVICE_NAME</txt-record>
     <txt-record>displayname=WENDY_DISPLAY_NAME</txt-record>
+    <txt-record>fqdn=WENDY_FQDN</txt-record>
```
Joannis: We're not using this, can we remove this?

kelenyibenjamin: Confirmed. Removed.

recipes-core/wendyos-identity/files/update-mdns-uuid.sh
```diff
@@ -51,43 +61,93 @@ set_txt() {
     sed -i -E "s|(<txt-record>$1=)[^<]*|\1$2|g" "$SERVICE_FILE"
 }
 
+# Validate a literal hostname as a single DNS label. Mirrors is_valid_hostname
+# in generate-hostname.sh and validHostname in the agent (services/hostname.go).
+is_valid_hostname() {
```
abularca: There is an identical function in `recipes-connectivity/avahi/files/generate-hostname.sh`, i.e `is_valid_hostname()`, which seems to be pretty similar to this one. I would extract this functionality in a common place so that they cannot get out of sync in time - _host name rule_.
On the other hand, the _device name rule_, implemented in two places:
1. `is_valid_device_name()` (recipes-core/wendyos-identity/files/generate-device-name.sh)
2. `validDeviceName()` (go/internal/agent/configpartition/apply.go)

they seem to disagree slightly with each other.  

To be clarified:
1. Do these two rules have to be somehow in sync?
2. Should we give the device name rule the trailing-hyphen check the hostcname rule already has?

Do all these make sense?

kelenyibenjamin: Thanks, both good points.

Hostname rule: extracted is_valid_hostname and read_name_file into a shared wendyos-identity-lib.sh that both update-mdns-uuid.sh and generate-hostname.sh now source, so there's a single source of truth.

Device-name rule:
* Yes, they should be in sync. They disagreed on length: the shell allowed up to 64 chars, the Go validDeviceName up to 55.  55 is the right cap (`wendyos-` + 55 = 63, the RFC 1035 label limit), so I tightened the shell to match.
*  Yes, added the trailing-hyphen reject to both. A trailing hyphen would otherwise derive wendyos-<name>-, an invalid label.

The Go validDeviceName change is a small. PR in the agent repo: https://github.com/wendylabsinc/WendyOS/pull/1874

abularca (approved):

PR #226, opened 2026-08-10
fix(tegra): expose NVIDIA EGL GBM platform to containers

Joannis:
## Problem

Containers on Jetson resolved EGL to **Mesa software rendering** instead of the Tegra driver. A Wayland kiosk (cage) on the HDMI output logged:

```
libEGL warning: MESA-LOADER: failed to retrieve device information
MESA: error: ZINK: vkCreateInstance failed (VK_ERROR_INCOMPATIBLE_DRIVER)
libEGL warning: egl: failed to create dri2 screen
```

followed by `EGLImage not supported` / `Failed to create FBO` and a laggy, CPU-rendered browser UI. Both failures are **silent** — EGL falls back rather than erroring, so nothing fails loudly.

## Root cause

meta-tegra's `drivers.csv` already maps the NVIDIA EGL/GLES userspace (`libEGL_nvidia.so.0`, `libnvidia-eglcore`, `libnvidia-egl-gbm`, `libnvidia-allocator`, the libglvnd vendor ICD) into containers — the `.so` files were present. Exactly two things were missing:

1. **`/usr/share/egl/egl_external_platform.d/15_nvidia_gbm.json` was never mapped.** `libEGL_nvidia.so.0` does not implement `EGL_PLATFORM_GBM_KHR` itself; it gains it by scanning that directory and `dlopen()`ing the library named there. With the JSON absent, the NVIDIA vendor advertises no GBM platform and libglvnd walks past `10-nvidia.json` onto `50_mesa.json`.

2. **GBM backends live at the wrong path for container images.** `drivers.csv` maps them as `/usr/lib/gbm/*_gbm.so` (Yocto layout). Debian/Ubuntu images build `libgbm.so.1` with `gbm-backends-path=/usr/lib/aarch64-linux-gnu/gbm`, so they never see them and fall through to Mesa's `dri_gbm.so`, which cannot drive the Tegra DRM node.

## Change

- `l4t.csv` + `l4t-blacksail.csv`: map `15_nvidia_gbm.json` and the three multiarch GBM backend symlinks.
- `nvidia-container-config_1.0.bb`: create host-side compat symlinks at `${libdir}/aarch64-linux-gnu/gbm/{nvidia-drm,tegra-udrm,tegra}_gbm.so` so `nvidia-ctk cdi generate` can mirror them in. Nothing on the host reads them — they exist purely as CDI source paths, hence `INSANE_SKIP:${PN} += "dev-so"` (bare `.so` links in a runtime package, dangling at package time because the target ships in meta-tegra's `tegra-libraries-*`).

The additions are heavily commented because they look redundant against `drivers.csv` and are easy to "tidy away" — the comments name what breaks if they're removed.

## Validation

Verified on hardware: **WendyOS 0.17.0 (blacksail, L4T R39.2, JetPack 7.2)** — the paths exist on the host and the kiosk renders GPU-composited.

Static checks in this branch:
- CSV record types are all valid (`dev`/`lib`/`sym`/`dir`); format matches existing conventions (`lib,` is already used for plain non-`.so` files elsewhere in these CSVs).
- `git diff --check` clean; both CSVs keep their trailing newline.
- Symlink resolution confirmed: `../../libnvidia-allocator.so` from `/usr/lib/aarch64-linux-gnu/gbm/` → `/usr/lib/libnvidia-allocator.so` → `.so.1` (mapped by `drivers.csv`), matching the recipe comment.

Not run locally: bitbake parse / packaging QA — no Yocto toolchain on macOS. The PR build (`meta-tegra-extensions/**` triggers the jetson family) is the real check, in particular that `INSANE_SKIP` covers the QA warnings these symlinks raise.

## Reviewer notes

- **`l4t.csv` (scarthgap/JP6) is unverified.** No JP6 device was reachable; the paths come from meta-tegra's `tegra-libraries-eglcore` / `egl-gbm` packaging, whose Yocto layout is identical on both trees. A CSV line whose host path does not exist is dropped by `nvidia-ctk cdi generate` with a warning, so the worst case is a no-op, not a broken container.
- **Side effect worth knowing:** `nvidia-container-cli` bind-mounts the *parent* directory of a mapped file, so containers see the host's `/usr/share/egl/egl_external_platform.d/` in place of their own. Today that dir holds only `15_nvidia_gbm.json`, so an image shipping its own `10_nvidia_wayland.json` would lose it — harmless here, since WendyOS doesn't ship `libnvidia-egl-wayland.so.1` on a mapped path either, so that platform was never usable in a container anyway.

https://claude.ai/code/session_018HAMgqsvSdRY9CSf1CWHwt

### DIFF
--- a/meta-tegra-extensions/recipes-core/packagegroups/packagegroup-nvidia-container.bb
+++ b/meta-tegra-extensions/recipes-core/packagegroups/packagegroup-nvidia-container.bb
@@ -76,12 +76,34 @@ RDEPENDS:${PN} = " \
     tegra-libraries-camera \
     tegra-libraries-eglcore \
     tegra-libraries-glescore \
+    egl-wayland \
     cudnn \
     cusparselt \
     tensorrt-core \
     tensorrt-plugins \
     libcufile \
     "
+
+# NOTE: egl-wayland is in that list for containers only, and it will look like a
+# mistake to anyone who greps DISTRO_FEATURES -- conf/distro/wendyos.conf removes
+# `wayland`, and nothing on the WendyOS host runs a compositor. It is here for the
+# same reason as everything else in this packagegroup: l4t.csv names
+# libnvidia-egl-wayland.so.1 and
+# /usr/share/egl/egl_external_platform.d/10_nvidia_wayland.json, and a CSV line
+# whose host path does not exist is silently dropped at CDI generation.
+#
+# What it buys: libEGL_nvidia.so.0 gains EGL_PLATFORM_WAYLAND_EXT, which is what
+# Wayland *clients* inside a container (GTK4, Qt, GPU-composited browsers) ask
+# for. Without it those clients fall through libglvnd to Mesa, which has no DRI
+# driver for the NVIDIA node, and land on llvmpipe -- a fully working picture
+# rendered on the CPU, so it reads as "the UI is laggy" rather than as an error.
+# The GBM entries in the same CSV only ever fixed the *compositor* side of this.
+#
+# The recipe (meta-tegra recipes-graphics/wayland/egl-wayland_git.bb) gates only
+# on REQUIRED_DISTRO_FEATURES = "opengl", which wendyos.conf adds for TensorRT.
+# Its wayland / wayland-protocols DEPENDS carry no REQUIRED_DISTRO_FEATURES of
+# their own and ship the -native variants it needs, so dropping the wayland
+# distro feature does not block this build.
 
 # DeepStream-specific packages (only when WENDYOS_DEEPSTREAM=1)
 # These provide libraries needed by DeepStream GStreamer plugins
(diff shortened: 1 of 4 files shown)


abularca (commented):
Not sure if it has an impact, but we will have to update the meta layers hashes, which is already prepared in the `feature/enable-tmp-data-encryption' branch, but some related packages have been upgraded:

|                                                            | SRCREV_TEGRA │ egl-wayland PV │ toolkit   |
| origin/main (EGL branch's base)       │ d0b2f4d8               │ 1.1.20                │ 1.16.2  |
| feature/enable-tmp-data-encryption  │ de0a5a23              │ 1.1.22+git          │ 1.19.1  |

Also, do we really need versioned wayland lib/sym? They look duplicated and maybe keeping only the json would be enough?
Aside from these questions, it looks fine to me.

PR #227, opened 2026-08-11
fix(tegra): size the Thor ESP image to its 512 MiB partition

Joannis:
## Problem

`wendy os update` on a Jetson AGX Thor fails with:

```
<3>wendyos-update: swap to slot A: stage capsule: write /boot/efi/EFI/UpdateCapsule/TEGRA_BL.Cap: no space left on device
```

after the full 3.6 GB download and a 130 s / 12 GiB rootfs slot write.

The ESP filesystem is 8x smaller than its partition. On `wendystudio-walter`:

```
/dev/nvme0n1p5   partition  536870912 B   (512 MiB)
                 filesystem 64511 KiB     (63 MiB)
```

Upstream `meta-tegra/recipes-bsp/uefi/tegra-espimage.bb` pins
`IMAGE_ROOTFS_SIZE` / `IMAGE_ROOTFS_MAXSIZE = 65536` (KiB), and
`image_types_tegra_esp.bbclass:oe_mkespfs` allocates exactly that
(`dd if=/dev/zero seek=${IMAGE_ROOTFS_SIZE} count=0 bs=1024`) before running
`mkfs.vfat -F 32 -I` over it. tegraflash writes that fixed-size image into
whatever the BSP layout declares for `esp`, and nothing ever grows the
filesystem afterwards — unlike `/data`, which `wendyos-data-setup` formats and
grows on first boot.

Thor is the only SOC where `wendyos-update` stages a UEFI capsule (staging is
gated on `tegra264`), and `tegra-bl.cap` is ~20.3 MiB. So a third of the usable
ESP had to be free for every bootloader OTA, and any space lost on that vfat was
unrecoverable in the field without a hand reformat.

## Change

Adds `WENDYOS_ESP_SIZE` (KiB) in a `tegra-espimage_%.bbappend`, defaulting to
`?= "65536"` — upstream's value — so the append is a no-op unless a machine opts
in. `conf/template/include/local/tegra-t264.inc` sets it to `524288` (512 MiB).

Scoped to t264 deliberately: t234's `esp` partition really is 64 MiB
(`67108864` in `flash_l4t_t234_nvme_rootfs_ab.xml`), so upstream is an exact fit
there and raising it would produce an image larger than its partition and break
flashing. Orin also never stages a capsule.

Third file is a comment fix — two places asserted a fixed 64 MiB ceiling that is
now a variable.

The `dd` is sparse and FAT32 is mostly zeros, so the compressed flash payload
barely changes.

## Testing

Not built or flashed — I have no build host here, so this is parse-correct by
inspection only. **Needs a Thor build + flash before merge.** Wiring verified by
hand:

- `conf/template/boards/jetson-agx-thor/local.conf:8` requires `local/tegra-t264.inc`
- `meta-tegra-extensions` is in Thor's BBLAYERS (`include/bblayers/tegra-base.inc:10`)
- that layer's `BBFILES` pattern `recipes-*/*/*.bbappend` matches the new file
- no other `tegra-espimage` bbappend exists in the tree
- bbappends parse after the `.bb`, so the strong `=` on `IMAGE_ROOTFS_SIZE` wins

On a built image, `df -k /boot/efi` on a freshly flashed Thor should report
~524288 total instead of 64511.

## Not covered here

This only helps devices flashed from a build carrying it. Thors already in the
field keep their 63 MiB ESP and need a manual reformat.

Three amplifiers remain open in `wendyos-update` (`internal/connector/tegrauefi/`),
which are what turned a tight ESP into a wasted 12 GiB write:

- no ESP free-space preflight in `PreflightInstall`, so the failure lands after
  the download and the rootfs write
- `copyFileSync` (`swap-slot.go:287`) writes the **final** path with
  `O_CREAT|O_TRUNC`, no temp+rename and no `os.Remove` on any error path — a
  short write destroys the previous capsule and leaves a truncated stub
- nothing ever sweeps `EFI/UpdateCapsule/`

https://claude.ai/code/session_01SmEDutXYdwGf7YFVSYLJQ8


abularca (approved):
Good findings, and the change is worth making, but not for the reason given.
A few things had to be fixed on the branch:
1. `tegra-espimage_%.bbappend` never applied. The upstream recipe is `tegra-espimage.bb`, unversioned, and BitBake's `%` only matches a <base>_ prefix. An unmatched append is an unconditional `bb.fatal` in this bitbake.
Renamed to `tegra-espimage.bbappend`
2. The causal story is not right, the ESP does not leak. A clean 63 MiB ESP holds three capsules and never accumulated (under normal circumstances), so routine operation cannot produce the ENOSPC you hit.
3. "Orin also never stages a capsule" is false.

Even though it is not clear why the board's ESP partition became full, the changes in the PR are correct and worth pushing. `wendyos-update` has been improved to report such an strange situation. It has now a new paramter `check` which verifies the ESP partition.

PR #228, opened 2026-08-11
Persist systemd-timesyncd's clock across A/B slots (WDY-1877)

thombles:
Closes WDY-1877.
Closes WDY-2197.
Pairs with [WendyOS#1683](https://github.com/wendylabsinc/WendyOS/pull/1683).

## What breaks for the user

An owner updates a Pi, or reboots it after a power cut. It comes back believing it is months in the past — the date the image was built. Log timestamps and update records are wrong, and if the owner has logged in to the CLI at any point since that date, the device rejects their credential as "not yet valid" and they are locked out of every channel until it reaches a time server. A Pi has no RTC, so this is every boot that starts without a network.

## What we have to work with

`systemd-timesyncd` already stamps `/var/lib/systemd/timesync/clock` on every sync and, at startup, advances the system clock to that file's mtime. That is exactly the "write the time down every time the device learns it" floor the ticket asks for, ticking continuously and needing no new code — but it lives on the A/B rootfs, so every OTA boots a slot whose copy came from the image and the value resets to the build date.

`/data` survives slot switches and already backs `/var/lib/wendy` and `/var/lib/bluetooth`. systemd's own fallback floor, `/usr/lib/clock-epoch`, is stamped 2011 (Yocto's `SOURCE_DATE_EPOCH`), so it is no help.

## Changes

- Bind-mount `/var/lib/systemd/timesync` from `/data/systemd-timesync`, ordered before `systemd-timesyncd` — it reads the file once at startup and is `Before=sysinit.target`, so a later mount would leave it on the rootfs copy.
- A prep unit creates the source with the ownership timesyncd needs and chowns it recursively. Yocto allocates system UIDs per image, so after a slot switch the clock file can carry the other slot's UID, leaving timesyncd unable to write it. The mount `Requires=` that unit.
- Drop-in ordering timesyncd after the mount, `Wants=` rather than `RequiresMountsFor=`: a failed mount should cost one boot on the rootfs copy, not disable time sync entirely.

This supersedes the ticket's proposal that the agent rewrite `/config/clock_floor` after each sync. The OS already maintains a ticking floor; making it survive an A/B switch is less code than duplicating it in the agent, and it applies before the agent starts, which is what WDY-2197's pre-NTP window needed.

## Validation (Pi 4, `wendyos-tom-rpi4`)

OTA from this PR's image on `p3` into `p4`, a real A/B switch:

```
System clock time advanced to recorded timestamp: Tue 2026-08-11 06:31:33 UTC
timesync: clock advanced ... offset 72.845s
```

72 seconds of residual drift against the 146-day jump measured before the change. All three units come up `active` from the shipped image.

The recursive chown was validated directly, since UID allocation happened to match across these two builds and cannot be forced from outside: a clock file planted at UID 991 with the units stopped was chowned and became writable as `systemd-timesync` once `Requires=` pulled the prep unit in. Without `Requires=`, a `systemctl restart systemd-timesyncd` brings the mount up with a `root:root` source and no chown — `x-systemd.mkdir` creates the bind source, not just the mount point — and the mount reads `active` either way, so it fails silently.

First boot is unchanged: the image ships `/var/lib/systemd/timesync/` empty (confirmed by mounting the rootfs underneath the bind mount), so there is no rootfs clock file being hidden and no seed value available.

## Not in this PR

WDY-1877 also floats triggering an immediate Roughtime re-query on network-up instead of the fixed backoff, which would shorten the stuck-clock window after a WiFi flap. Untouched here.

Seeding the persisted clock from a real build timestamp would give a permanently-offline RTC-less device a usable floor instead of 2011. Worth its own issue.

### DIFF
--- a/recipes-core/systemd-mount-wendy/files/timesyncd-persist-clock.conf
+++ b/recipes-core/systemd-mount-wendy/files/timesyncd-persist-clock.conf
@@ -1,14 +1,12 @@
 [Unit]
-# timesyncd reads /var/lib/systemd/timesync/clock once, at startup, and it is
-# Before=sysinit.target with no ordering against local-fs.target — so without
-# this it can start before the bind mount and read the rootfs copy instead of the
-# persisted one.
+# timesyncd reads /var/lib/systemd/timesync/clock once, at startup, and is
+# Before=sysinit.target with no ordering against local-fs.target — so without this
+# it can start before the bind mount and read the rootfs copy.
 #
-# Wants=, NOT RequiresMountsFor= — the opposite choice to
-# bluetooth-persist-state.conf, deliberately. There, a failed mount should stop
-# bluetoothd rather than let it come up amnesiac. Here, blocking timesyncd on a
-# failed mount would leave the clock unsynced, which is the exact failure this
-# change exists to prevent. Failing open costs one boot on the rootfs copy —
-# i.e. the behaviour before this change — instead of no time sync at all.
+# Wants=, not RequiresMountsFor= — the opposite choice to
+# bluetooth-persist-state.conf. There a failed mount should stop bluetoothd rather
+# than let it come up amnesiac; here blocking timesyncd would leave the clock
+# unsynced, the failure this change exists to prevent. Failing open costs one boot
+# on the rootfs copy, which is the behaviour before this change.
 Wants=var-lib-systemd-timesync.mount
 After=var-lib-systemd-timesync.mount
--- a/recipes-core/systemd-mount-wendy/files/timesyncd-persist-clock.conf
+++ b/recipes-core/systemd-mount-wendy/files/timesyncd-persist-clock.conf
@@ -1,0 +1,14 @@
+[Unit]
+# timesyncd reads /var/lib/systemd/timesync/clock once, at startup, and it is
+# Before=sysinit.target with no ordering against local-fs.target — so without
+# this it can start before the bind mount and read the rootfs copy instead of the
+# persisted one.
+#
+# Wants=, NOT RequiresMountsFor= — the opposite choice to
+# bluetooth-persist-state.conf, deliberately. There, a failed mount should stop
+# bluetoothd rather than let it come up amnesiac. Here, blocking timesyncd on a
+# failed mount would leave the clock unsynced, which is the exact failure this
+# change exists to prevent. Failing open costs one boot on the rootfs copy —
+# i.e. the behaviour before this change — instead of no time sync at all.
+Wants=var-lib-systemd-timesync.mount
+After=var-lib-systemd-timesync.mount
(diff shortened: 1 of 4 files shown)


thombles:
## Hardware validation round 1 — found a real defect, fixed in c934c7f

OTA'd a Pi 4 to this PR (`0.18.2` p4 → `pr-228` p3; the "likely rolled back" warning is the known WDY-2196 false positive, the slot demonstrably switched).

**Ordering: confirmed good.** Boot journal, monotonic:

```
[ 8.170] Starting Prepare persistent systemd-timesyncd state directory...
[ 8.248] Finished  Prepare persistent systemd-timesyncd state directory.
[ 8.452] Mounted   Bind mount /var/lib/systemd/timesync ...
[ 8.503] Starting  Network Time Synchronization...    <- after the mount
```

No ordering cycles, no failed units. Dropping `After=grow-data-fs-online.service` was necessary and sufficient.

**But the clock still came up at the build epoch.** The mount was active and ordered correctly, and it still didn't work:

```
[21.06] timesync: clock advanced from=1773416125 to=1786062782 offset=12646656
```

That is the *agent* jumping 146 days from the Yocto build epoch to the install-time `clock_floor` — i.e. timesyncd never restored anything.

### Cause: system UIDs differ between the two slot images

```
this slot systemd-timesync uid/gid: 991/988
clock file numeric owner:           993/990  ->  rtkit/empower on this slot
touch: cannot touch '/data/systemd-timesync/clock': Permission denied
```

Yocto allocates system UIDs dynamically, so `systemd-timesync` is **991 in one slot's image and 993 in the other**. `/data` is shared, so after the switch the clock file carries the other slot's UID, timesyncd cannot write it, and it then *silently* declines to restore the clock — no log line at all. The device behaves exactly as if this mount were absent.

### Fix (c934c7f) and confirmation

Recursive chown in the prep service, so whichever UID the running slot allocated becomes the owner. Verified by hand on the device before committing — chown the file, reboot the same slot:

| | before chown | after chown |
| -- | -- | -- |
| clock at 25s | build epoch | `1786419819` (correct) |
| `clock advanced from 1773416125` | present | **absent** |
| agent correction | two jumps, 146 days | single +50s |

Note `FIRST ENTRY` in `journalctl --list-boots` stays at the build epoch either way — journald flushes kmsg before the bump, so it is not a valid discriminator. The `clock advanced from <build epoch>` line is.

### Broader implication worth a separate look

**Any `/data` state owned by a service user has this bug.** `/var/lib/bluetooth` escapes it only because `wendyos-bluetooth-state.service` uses `root:root` — UID 0 is the one UID that cannot drift. Pinning system UIDs distro-wide (`useradd-staticids`) would be the systemic fix.

Re-running CI, then a second OTA to validate end-to-end across a real slot switch. Checklist item "a second OTA comes up at ~last-sync time" is still open — that is what round 2 tests.

PR #230, opened 2026-08-13
base-files: enable TCP MTU probing so a carrier uplink cannot black-hole the agent

chrisdok43:
Fixes the root cause in [WDY-2443](https://linear.app/wendylabsinc/issue/WDY-2443/wendyos-pmtu-black-hole-on-lte-uplinks-silently-kills-the-agents-wendy).

## The failure

A device whose default route is an LTE modem **goes dark to Wendy Cloud while still having working internet**. It vanishes from `wendy cloud discover`, `wendy cloud device shell` fails, cameras stop reaching the relay — yet `wendyos-agent` is running, the link is up, DHCP has a lease, and ping/DNS/traceroute all succeed.

The carrier path's MTU is under 1500 and it drops the ICMP "fragmentation needed" replies PMTUD depends on, so PMTUD fails **silently**: small packets pass, large ones vanish, nothing errors.

TCP connects fine. The first large packet is a TLS ClientHello — ~1556 bytes, because a TLS 1.3 post-quantum key share is over 1.2 KB on its own — and it is never delivered:

```
* Trying 104.26.12.205:443...
* socket successfully bound to interface 'eth0'    <- TCP handshake SUCCEEDS
* TLSv1.3 (OUT), TLS handshake, Client hello (1):
} [1556 bytes data]                                 <- first LARGE packet
* Connection timed out after 15005 milliseconds
```

Everything the agent does to reach the cloud is TLS, so all of it fails, in a 90s retry loop that never converges.

**It does not take an LTE-only site to hit this.** `99-interface-metrics.conf` gives ethernet metric 100 against WiFi's 300, so plugging an LTE modem into `eth0` moves cloud egress onto it *even with WiFi up and healthy*. LAN/mDNS management keeps working, so the device looks fine to anyone on site and dead to everyone else. The whole failure was reproduced with WiFi connected throughout.

## The change

One shipped sysctl: `net.ipv4.tcp_mtu_probing = 1`.

The kernel falls back to packetization-layer PMTUD when it detects a black hole and searches for a working MSS, instead of retransmitting into the void. It **adapts to whatever the carrier's number is** rather than hardcoding one that goes stale on a different SIM, APN or site.

Why `1` and not `2`: `1` engages only after a black hole is detected — one stalled connection on a broken path, and **nothing at all on a healthy link**, which is what makes it safe fleet-wide. `2` probes from the start and lowers the initial MSS on every link, including the LANs where nothing is wrong.

**Why it ships in the image rather than being applied to devices:** `/etc/sysctl.d` belongs to the running A/B slot, so a hand-applied drop-in survives a reboot but is lost on the next OS update — exactly when nobody is watching. Baking it in means every slot has it.

**Scope — TCP only.** The UDP/QUIC half would need an MTU clamp, deliberately *not* done fleet-wide: the safe carrier value (~1400) would also shrink every packet on the gigabit LANs where nothing is wrong. The recipe comment and the troubleshooting entry give the per-connection `nmcli` clamp for sites known to be behind a carrier link.

## Evidence

Measured on a Jetson Orin Nano on a TELUS LTE modem, stepping the link MTU and probing with a real HTTPS request:

| eth0 link MTU | 1500 | 1480 | 1460 | 1440 | 1434 | 1430 | 1428 | 1420 | 1400 | 1280 |
|---|---|---|---|---|---|---|---|---|---|---|
| HTTPS over eth0 | ✗ | ✗ | ✗ | ✗ | ✗ | ✓ | ✓ | ✓ | ✓ | ✓ |

**Usable path MTU 1430–1433.** With `tcp_mtu_probing=1`, HTTPS succeeds at MTU **1500** in ~4s, three attempts of three.

Clamping the MTU by hand brought the device back into `wendy cloud discover` **within one poll and with no agent restart** — it self-heals on its own retry the moment packets get through. The agent was never at fault.

## Verification

I can't run a Yocto build here, so I verified the delivery mechanism on a live WendyOS 0.18.1 device (ccr1) instead:

```
/etc/sysctl.d/  →  99-quiet-console.conf, 99-usb-network-performance.conf
kernel.printk   =  3 4 1 3      ← the shipped drop-in IS applied at runtime
net.ipv4.tcp_mtu_probing = 0    ← currently off, confirming the gap
```

So a `base-files` drop-in reaches `/etc/sysctl.d` and is applied by `systemd-sysctl` on this image. This change follows `99-quiet-console.conf` exactly — same recipe, same install pattern. `base-files` rather than `meta-tegra-extensions` because the problem is board-agnostic: any device on a carrier uplink hits it. Nothing else in the repo sets `tcp_mtu_probing`, so there's no conflict.

**What still needs a real build/boot:** that the file lands in the image and reads back as `1`. Worth confirming on the first image, plus the `nmcli radio wifi off` A/B and reboot survival on the affected device.

## Also included

A troubleshooting entry (`docs/jetson-troubleshooting.md`) with symptom → cause → fix → verify, because this failure is nearly undiagnosable from the device's own vantage point. It records two traps that each cost real time:

- **`ping -M do` cannot measure PMTU on this image.** BusyBox `ping` has no `-M` flag, so every rung returns "invalid option" and reads as a failure — *including over a healthy link*, which looks like a catastrophic MTU problem and sends you the wrong way. Probe with `curl --interface <iface>` while stepping the link MTU.
- `nc -z` misreports here too; it failed against endpoints `curl` reached over the same interface.

### DIFF
--- a/docs/jetson-troubleshooting.md
+++ b/docs/jetson-troubleshooting.md
@@ -121,3 +121,81 @@ slot: 1,    retry_count: 0,    status: normal
   OTA cycle, it is safe to delete: `rm /data/mender/tegra-bl-version-before`
 
 ---
+
+## Device is invisible to Wendy Cloud on an LTE / carrier uplink
+
+### Symptom
+
+The device vanishes from `wendy cloud discover` and `wendy cloud device shell`
+fails — but it is **not** offline. `wendyos-agent` is running, the link is up,
+DHCP has a lease, ping and traceroute succeed, and other software on the box
+keeps using the network normally. The agent retries every 90s forever:
+
+```
+broker connection failed, reconnecting  ... authentication handshake failed: EOF  backoff:90
+cloud flusher: flush failed             ... authentication handshake failed: EOF
+mesh roster sync failed                 ... DeadlineExceeded
+```
+
+It does **not** take an LTE-only site to hit this. `99-interface-metrics.conf`
+gives ethernet metric 100 against WiFi's 300, so plugging an LTE modem into
+`eth0` moves cloud egress onto it even with WiFi up and healthy — LAN/mDNS
+management keeps working, which is why the device looks fine to anyone on site
+and dead to everyone else.
+
+### Cause
+
+The carrier path's MTU is below 1500 and it drops the ICMP "fragmentation
+needed" replies Path MTU Discovery depends on. PMTUD fails silently: small
+packets pass, large ones are black-holed with no error.
+
+TCP connects fine. The first *large* packet is a TLS ClientHello — ~1556 bytes,
+since a TLS 1.3 post-quantum key share is over 1.2 KB by itself — and it is
+never delivered, so every TLS connection hangs until it times out.
+
+### Fix
+
+Shipped since this was found: `/etc/sysctl.d/99-wendyos-pmtu.conf` sets
+`net.ipv4.tcp_mtu_probing = 1`, so the kernel searches for a working MSS when it
+detects a black hole. On an image that predates it, apply by hand:
+
+```bash
+printf 'net.ipv4.tcp_mtu_probing = 1\n' > /etc/sysctl.d/99-wendy-pmtu.conf
+sysctl -p /etc/sysctl.d/99-wendy-pmtu.conf
+```
+
+That drop-in lives on the running A/B slot and is **lost on the next OS update** —
+which is why the real fix ships in the image. For UDP/QUIC, which probing does
+not cover, clamp the connection on sites known to be behind a carrier link:
+
+```bash
+nmcli connection modify "<profile>" 802-3-ethernet.mtu 1400
+nmcli connection down "<profile>" && nmcli connection up "<profile>"
+```
+
+### Verify
+
+```bash
+sysctl net.ipv4.tcp_mtu_probing          # expect 1
+curl -sS --interface eth0 --max-time 20 -o /dev/null -w '%{http_code}\n' https://api.ipify.org
+```
+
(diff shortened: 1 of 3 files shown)


kelenyibenjamin:
Duplicate work: PRs https://github.com/wendylabsinc/WendyOS-Builder/pull/234 and https://github.com/wendylabsinc/WendyOS-Builder/pull/230 both create recipes-core/base-files/files/sysctl.d/99-wendyos-pmtu.conf and both edit base-files_%.bbappend. Worth de-duplicating those before landing anything.

EBro912:
Heads up: the sysctl mechanism this PR ships (`tcp_mtu_probing=1` + `tcp_base_mss=1200`) was verified on-hardware on 2026-08-25 against a real LTE/CGNAT black hole (Orin Nano + MikroTik KNOT LTE4, path MTU 1460 with frag-needed silently dropped) — it rescued the agent's broker handshake on the first dial attempt (~8s after restart) where stock settings loop on `authentication handshake failed: EOF` indefinitely. Full methodology and results are in the comment on #234, which ships the same sysctl plus an NM-level MTU clamp — the two PRs overlap and probably want reconciling.

One testing note that applies to this PR's verification steps too: emulating the black hole with device-local netfilter length drops gives false failures (the OUTPUT hook sees pre-GSO aggregate skbs, which stay oversized after MSS reduction) — build the hole router-side instead.

justsem:
Superseded by #234, which now ships exactly the tcp_mtu_probing=1 sysctl (carrier-specific MTU/MSS tuning stays on the LTE modem, per RouterOS MSS clamp + DHCP option 26 notes in #234).

PR #231, opened 2026-08-14
fix: s5cmd checksums asset URL in install-build-deps.sh

EBro912:
Cherry-pick of a fix discovered while build-verifying the WDY-2430 v4l2loopback branch — filed separately because it unblocks `make setup` for **every** board, not just that feature.

The s5cmd v2.2.2 GitHub release publishes its checksums file **unversioned** (`s5cmd_checksums.txt`); the script requested a version-prefixed name and 404'd, so bootstrap failed on any fresh machine (bug present on `main` since `b04163a`). The per-file entries inside the checksums file are version-prefixed, so the existing `awk` lookup is unchanged.

Verified against the live GitHub release assets, and exercised end-to-end by the (previously impossible) `make setup` + `bitbake v4l2loopback` run on the feature branch.

Recommend merging this before the `wdy-2430-v4l2loopback-module` PR (which carries the same commit; it will dedupe on merge).


kelenyibenjamin:
PR closed as #232 already contains this change.

PR #233, opened 2026-08-14
ci: warm only when a compile is needed, and build variants in parallel

kelenyibenjamin:
`bitbake -n` now runs first. It executes nothing and writes no stamps, but lists the tasks a real build would run. If none of them compiles, the warm is skipped and the matrix restores from sstate as it already would have. Any probe failure warms, so the fallback is today's behaviour.

One job per variant again, so they queue independently instead of running sequentially inside one job. The warm matrix stays merged, so the MACHINE-independent bulk is still compiled once per device, not once per disk.

`warm-caches` is dispatched by a paths filter, but whether a warm is needed depends on task hashes. A one-line edit to `conf/distro/wendyos.conf` trips the `global` filter and warms all 7 devices.

`publish` now holds a device back until every one of its variants has an entry (previously a failed SD build could bump the device's `latest` to a version with no SD image), and the agent version is resolved once per run instead of once per job, so two variants of one board can't pin different agents under one version.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -583,6 +583,64 @@ jobs:
             echo "WENDYOS_AGENT_SHA256=$sha"
           } >> "$GITHUB_ENV"
 
+      # The paths filter only knows that a file under a heavy directory changed.
+      # A dry run lists the tasks a real build would execute without running or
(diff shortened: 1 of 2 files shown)


kelenyibenjamin:
Draft. No review needed. Testing the CI.

PR #234, opened 2026-08-14
fix(network): recover from PMTU black holes on LTE/CGNAT uplinks

EBro912:
## Summary

Fixes [WDY-2443](https://linear.app/wendylabsinc/issue/WDY-2443): a WendyOS device whose uplink is a carrier LTE/CGNAT link (path MTU ~1430, ICMP frag-needed dropped) goes dark to Wendy Cloud while looking perfectly healthy. TCP connects, but the agent's ~1556-byte post-quantum TLS ClientHello rides in a full-MSS segment that exceeds the path MTU and is silently black-holed, so every TLS handshake hangs and the agent retries forever. Reproduced end to end on Enmax1 (Jetson Orin Nano, TELUS LTE).

This PR is now scoped to **RFC 4821 packetization-layer PMTU discovery only** — the kernel infers the real path MTU from the retransmission pattern, needing no ICMP feedback. The fix ships in the image because `/etc/sysctl.d` is not persisted from `/data` — an on-device fix would silently vanish on the next A/B OTA.

## Changes

- **`recipes-core/base-files`**: new sysctl drop-in `99-wendyos-pmtu.conf`:
  - `net.ipv4.tcp_mtu_probing=1` — arm the black-hole fallback (probe only once a black hole is detected, after ~3s of failed retransmission). Verified rescuing HTTPS at link MTU 1500 on the affected device (~4s recovery).
  - `net.ipv4.tcp_base_mss=1200` — first fallback step; fits any path down to the 1280-byte IPv6 minimum, so one step recovers essentially every real constriction.
  - `net.ipv4.tcp_mtu_probe_floor=512` — bounds a false positive (ordinary loss looks like a black hole for ~3s) at a slow-but-usable segment size instead of the kernel-default 48.

  Installed via `base-files` so it reaches every machine (Tegra/RPi/x86/QEMU). Named to sort after the Tegra `99-usb-network-performance.conf` so a future key overlap can't override it.

## What was removed, and where the rest of the fix lives

These sysctls are per-netns and TCP-only: containers, UDP/QUIC, and the inbound direction (advertised MSS comes from the link MTU) are still exposed. Earlier revisions of this PR closed those gaps host-side — first a fleet-wide NM `ethernet.mtu=1400` connection default, then a socket-evidence-driven runtime clamp (`uplink-mtu-tuning`). Both are now dropped: the chosen fix for those gaps is the LTE modem itself (MikroTik KNOT, RouterOS: MSS clamp in the firewall mangle + DHCP option 26), which is deterministic at connection setup and also reaches container inbound traffic, which no host-side setting can. The sysctls remain as the safety net for links we don't own (customer routers, phone hotspots, cellular failover added after we ship).

Also reverted with the clamp: the `systemd-conf` networkd changes (dhcp-ethernet PACKAGECONFIG removal + wait-online mask) and the MOTD probe update. They entered this PR only so a runtime MTU change would stick. The dual-link-management problems they fixed (two DHCP leases and two default routes on every ethernet link, `systemd-networkd-wait-online` burning its full 119s timeout and delaying `network-online.target`) are real but independent — they should get their own PR.

## Verification

- Recipe change mirrors the existing install patterns line-for-line; not yet image-build-verified.
- Suggested: build `qemuarm64-wendyos` + `jetson-orin-nano-devkit-wendyos`, confirm the file lands and `sysctl net.ipv4.tcp_mtu_probing` = 1 in the booted image.
- On-device acceptance on Enmax1 (WiFi off): appears in `wendy cloud discover`, `wendy cloud device shell` works, survives reboot **and** an A/B OTA. Note Enmax1's current MTU 1430 is runtime-only — reboot it first to restore the broken baseline for an honest test.


EBro912:
**On-hardware verification of both halves (runtime mechanisms), 2026-08-25** — Orin Nano on stock WendyOS 0.18.2 (`tcp_mtu_probing=0` confirmed), real LTE/CGNAT uplink (MikroTik KNOT LTE4 / EG25-G, EiotClub SIM roaming on Verizon), agent carrying the wendylabsinc/WendyOS#1709 diagnostic.

Two site facts worth having in the record, since they explain why a device can look immune until it isn't:
1. The carrier path here measured true MTU **1460 with frag-needed silently dropped** — a real black hole — yet the stock agent still reached the broker, because **Google's front end clamps MSS to 1400** (1452-byte frames squeak under 1460). Enmax1's ~1430 sits *below* that clamp, which is exactly why it went dark. Any path under ~1452, or any non-clamping endpoint, breaks.
2. With a global IPv6 on WiFi, the broker dial prefers AAAA and **escapes the LTE path entirely** — reproducing the failure requires killing wlan0 IPv6, not just relying on route metrics (v4 metrics already prefer ethernet).

Test method: wlan0 IPv6 off (runtime), then a genuine on-wire hole: router LTE-side MTU **1420** + firewall drop of the router's outgoing ICMP frag-needed. At 1420, Google-clamped 1452-byte frames die while a 1252-byte frame (base_mss 1200) and 1400-clamped frames pass — so each mechanism is separable. Baseline under the hole: infinite `authentication handshake failed: EOF` loop, backoff 1→64, absent from `wendy cloud discover`.

- **Sysctl half** (`tcp_mtu_probing=1`, `tcp_base_mss=1200`, applied runtime): presence registered **8 seconds after agent restart, first dial attempt**; device reappeared in `wendy cloud discover`.
- **Clamp half alone** (`ip link set eth0 mtu 1400`, sysctls stock): presence registered in **3 seconds**. Each half is independently sufficient on this failure profile.

Verifier's warning: do **not** emulate the hole with device-local netfilter length drops — the OUTPUT hook sees pre-GSO aggregate skbs and keeps dropping after MSS reduction, which falsely "disproves" the sysctl half (it cost a test cycle here). Router-side MTU + ICMP suppression is faithful.

Scope: runtime mechanisms only. Image-build persistence (files landing, surviving reboot + A/B OTA) still needs the Builder image acceptance pass described in the PR.

PR #235, opened 2026-08-14
fix(tegra): re-enable the TCP metrics cache

EBro912:
## Summary

Part of [WDY-2443](https://linear.app/wendylabsinc/issue/WDY-2443), split out so it can be discussed independently of the outage fix (#234).

`net.ipv4.tcp_no_metrics_save=1` in the Tegra-only `99-usb-network-performance.conf` discards learned per-destination TCP metrics — including the discovered path MTU — after every connection. On LTE/CGNAT uplinks whose path black-holes full-size packets, that makes every new connection rediscover the black hole from scratch instead of reusing what the previous connection learned via `tcp_mtu_probing`. And Tegra is exactly the hardware deployed behind LTE.

This restores the kernel default (0), keeping the key explicit with a comment explaining why.

If the original setting was intentional for USB-gadget throughput benchmark repeatability, it should be applied ad hoc during benchmarks rather than shipped fleet-wide — happy to discuss.

### DIFF
--- a/meta-tegra-extensions/recipes-support/usb-network-tuning/files/99-usb-network-performance.conf
+++ b/meta-tegra-extensions/recipes-support/usb-network-tuning/files/99-usb-network-performance.conf
@@ -42,8 +42,12 @@ net.ipv4.tcp_timestamps = 1
 # Enable selective acknowledgments for better recovery from packet loss
 net.ipv4.tcp_sack = 1
 
-# Don't cache TCP metrics from previous connections
-net.ipv4.tcp_no_metrics_save = 1
+# Cache TCP metrics from previous connections (kernel default). The cache
+# carries learned path-MTU/MSS across connections to the same peer, which
+# matters on LTE/CGNAT uplinks where PMTU discovery is black-holed
+# (WDY-2443): with the cache disabled, every new connection rediscovers
+# the black hole from scratch.
+net.ipv4.tcp_no_metrics_save = 0
 
 # TCP memory allocation: min, pressure, max (in pages)
 net.ipv4.tcp_mem = 786432 1048576 26777216


kelenyibenjamin (commented):
the value change is right - 0 is the kernel default, and the = 1 came from: https://github.com/wendylabsinc/WendyOS-Builder/commit/7593d3b0#diff-04b6c61468c44ed0002b6e48f0dd375f53423f311d60b861c374dfa39723ce57R46, but the comment is wrong:   the tcp_metrics cache stores no path MTU (checked on my local tree: 6.8.12)

Duplicate work: PRs #234 and #230 both create recipes-core/base-files/files/sysctl.d/99-wendyos-pmtu.conf and both edit base-files_%.bbappend. Worth de-duplicating those before landing anything.

cc: @abularca

EBro912:
**Confirmed on hardware, 2026-08-25**: an Orin Nano on stock WendyOS 0.18.2 ships `net.ipv4.tcp_no_metrics_save = 1` (read on-device), so learned per-destination TCP state is discarded on every connection close today, exactly as this PR describes — and Tegra-behind-LTE is indeed the deployment where that hurts.

During the WDY-2443 runtime testing (setup in the wendylabsinc/WendyOS-Builder#234 comment), with this PR's value (`0`) applied, `ip tcp_metrics show` retained per-destination entries (cwnd, rtt/rttvar) for all four broker IPs, keyed to the LTE source address, after a PLPMTUD-rescued connection — the cache the reconnect loop would warm-start from. With the stock `1`, the table stays empty. This pass didn't isolate a before/after convergence-time delta for the cached PMTU specifically (the #234 sysctls already rescued the first attempt in ~8s), so the quantitative benefit on reconnect is inferred from mechanism rather than measured here.

No adverse effects observed on the USB-gadget path during the session (USB-C NCM control connection stayed healthy throughout).

PR #236, opened 2026-08-14
driver add-ons: install kernel modules without reflashing

kelenyibenjamin:
Ships kernel driver add-ons as systemd-sysext images, so a device can gain a driver without an OS rebuild or a reflash.

The idea: each build publishes a **kernel devkit**,  a relocatable copy of the kernel build tree plus its toolchain, keyed on `uname -r`. Drivers compile against that devkit instead of against a Yocto tree, so adding one never rebuilds a kernel and never touches the self-hosted builder.

    OS build   ->  devkit for <uname -r>   ->  manifest
    driver PR  ->  build .ko + pack .raw   ->  manifest extensions
    device     ->  /data/extensions/enabled ->  merged onto /usr at boot

Add-ons live on `/data`, so they survive an OTA.

Enabled on `raspberrypi5-wendyos` only, via `WENDYOS_DRIVER_EXTENSIONS`.


kelenyibenjamin:
After discussing this with Adrian, we agreed that driver add-ons should be kept separate whenever possible.
Thanks to @abularca for the commit that includes the discussed optimizations.

recipes-core/wendyos-sysext-apply/files/wendyos-sysext-apply.service
```diff
@@ -0,0 +1,17 @@
+[Unit]
+Description=WendyOS: merge driver add-ons from /data (sysext + depmod)
+Documentation=man:systemd-sysext(8)
+# Ordered on data.mount alone, deliberately not on grow-data-fs-online.service: /data is
+# safe to read and write during an online resize, which can take minutes on a large card,
+# so waiting for it would only delay every driver.
+After=data.mount
+Wants=data.mount
```
abularca: `Wants=` deliberately tolerates a failure, so I would change the service file this way:
```
RequiresMountsFor=/data
After=data.mount
```

I would also remove `ConditionPathExists=/data` because my assumption is that a mount-point condition would turn a broken `/data` into a green unit with no add-ons loaded - silent no-op.

abularca (approved):
Well done!

PR #237, opened 2026-08-14
ci: retire warm-caches by pinning bitbake parallelism

kelenyibenjamin:
`BB_NUMBER_THREADS` and `PARALLEL_MAKE` never reached bitbake, the Makefile narrows the env passthrough and drops them, so both fell back to nproc (64). Every job was running **64 tasks × `-j 64`** inside a 16 GiB cgroup, which is what made compile-heavy builds thrash and time out.


conf/template/include/local/common.inc
```diff
@@ -34,6 +34,17 @@ BB_DISKMON_DIRS ??= "\
     HALT,${SSTATE_DIR},100M,1K \
     HALT,/tmp,10M,1K"
 
+# Bitbake parallelism, pinned in-conf because the build Makefile narrows the env
+# passthrough and neither variable survives it; unset, both fall back to nproc
+# (64 on the builder). 64 tasks x -j 64 is what made a compile-heavy build unfit
+# for the 16 GiB job cgroup — at 8 x 8 a matrix job builds a kernel unaided.
+BB_NUMBER_THREADS = "8"
```
abularca: These 3 new assignments need to be `?=` instead of `=`, and specifically not `??=`.
`=` blocks the override route the rest of this file relies on (e.g. it will override any value in `auto.conf`)
`??=` looks like the safe choice and is actually a silent no-op.

kelenyibenjamin: Thanks @abularca  to highlight this. I double checked and  you are totally right. Changes done.

justsem:
LGTM. One thing to keep an eye on: The CI infrastructure we have does dynamic admission based on current load.
using `-j 8` certainly influences that and might lead to more parallelism but it might also increase build times, so let's see what works.

PR #240, opened 2026-08-20
Remove scarthgap support

abularca:
Every board but qemu-arm64 already built on blacksail, and oe-core declares:
 `LAYERSERIES_CORENAMES = "wrynose blacksail"`, 
so every scarthgap gate was already false. 
These have been removed:
- the four parked rpi*-scarthgap templates
- the jp6 layer
- the paired scarthgap/modern includes
- the 6.6-only kernel patches were all unreachable


meta-qemu-extensions/recipes-kernel/linux/linux-yocto_%.bbappend
```diff
@@ -8,17 +8,16 @@ KMACHINE:qemuarm64-wendyos = "qemuarm64"
 
 FILESEXTRAPATHS:prepend := "${THISDIR}/${PN}:"
```
kelenyibenjamin: Is this needed? It seems that FILESEXTRAPATHS:prepend now points at a directory that no longer exists after all 7 files under it were deleted.

meta-x86-extensions/conf/layer.conf
```diff
@@ -11,15 +11,14 @@ BBFILE_PRIORITY_wendyos-x86 = "51"
 
 # x86 .wks layouts (A/B) live in files/wic/ — the location blacksail oe-core
 # mandates (its sanity checker rejects the older <layer>/wic, and its
-# WKS_SEARCH_PATH only searches <layer>/files/wic). Mirrors meta-rpi-extensions.
+# WKS_SEARCH_PATH only searches <layer>/files/wic). The append states that path
+# outright so wic finds the layouts no matter how the search path is built.
 WKS_SEARCH_PATH:append = ":${LAYERDIR}/files/wic"
```
kelenyibenjamin: Is there a reason why WKS_SEARCH_PATH:append was keep here?
I observed that in  meta-rpi-extensions was deleted.

abularca: No good reason, removed it here too, so the two layers match.

meta-qemu-extensions/recipes-kernel/linux/linux-yocto_%.bbappend
abularca: Good catch! The linux-yocto/ directory went away.

kelenyibenjamin (approved):

PR #241, opened 2026-08-24
Ship BuildKit, so a build host survives an OS update

chrisdok43:
**Draft: never been through bitbake.** See "What is unverified" — this needs an image build before it can be trusted.

Pairs with WendyOS #1767, which makes the same hazards visible from the CLI.

## The problem

`wendy run --build-host <device>` needs buildkitd on the device, and nothing provides it. Installing it by hand works — until the next OS update.

Verified on `joannis-agx-thor`:

| path | on | survives an OS update? |
|---|---|---|
| `/usr/local/bin` | `/` (A/B rootfs slot) | **no** |
| `/etc/systemd/system` | `/` (A/B rootfs slot) | **no** |
| `/data` | `nvme0n1p17`, 912 GB | yes |

So an update boots a slot with neither the daemon nor its unit. **And the failure is quiet:** the agent's opt-in marker lives at `/data/etc/wendy-agent/build-host-enabled` and *does* survive, so the device keeps reporting the builder role as enabled while having nothing to build with — a state nobody can explain from the outside.

## Second problem, same cause: where the cache goes

`buildkitd`'s default `--root` is `/var/lib/buildkit`, which on that Thor is a **12 GB filesystem with 4.6 GB free**, beside 862 GB of `/data`. A build cache grows by design, and an edge image that compiles a TensorRT engine writes gigabytes — so the default does not produce a failed build, it fills the partition the device boots from. theta already sits at 97% on `/`.

`buildkitd.toml` therefore sets `root = "/data/buildkit"`, plus a GC policy — because that file is all that stands between the two outcomes if `root` is ever changed.

## Deliberate choices

**Off by default** (`WENDYOS_BUILD_HOST = "0"`). ~85 MB most devices will never use, and a camera box has no business running other people's builds. Opt in on the Sparks and AGX Thors.

**Installed but not auto-enabled.** The agent already gates `BuildImage` on its own marker file, so the unit ships disabled and enabling the role starts it. Starting a daemon on every image would spend memory on hosts that never build.

**Binaries in `/opt/buildkit/bin`, symlinks on PATH.** For the reason `wendyos-agent` already documents: a merged sysext driver add-on turns `/usr` into an ephemeral overlay, so anything installed there is discarded on reboot. The agent execs a bare `"buildctl"` (`build_service.go`), so the symlinks are load-bearing, not cosmetic.

**containerd worker disabled.** containerd on these devices belongs to the agent that runs the apps; a second client sharing its content store and namespaces means two writers to state the agent believes it owns.

**`OOMScoreAdjust=500`** — a build is best-effort work on a device whose real job is running apps, so it should die before the agent does.

**`RequiresMountsFor=/data`** — a daemon that starts before `/data` is mounted would create its state directory on the root filesystem, which is the exact failure this configuration exists to avoid.

## What is verified, and what is not

**Verified, by hand, on real hardware:** buildkit 0.32.2 with these binaries on a Jetson AGX Thor and a DGX Spark, both building and delivering images to Jetson camera boxes in another country. `root` on `/data` confirmed working; the default confirmed dangerous (4.6 GiB free).

**Not verified:** this recipe. It has never run through bitbake. Specifically unchecked —

- whether the prebuilt-binary inhibits (`INHIBIT_PACKAGE_STRIP` and friends) are the right set for this tarball
- whether `COMPATIBLE_MACHINE` and the arch mapping behave on an x86 machine
- image size impact with the flag on
- that `SYSTEMD_AUTO_ENABLE = "disable"` produces an installed-but-inactive unit as intended
- that the `find`-based install survives an upstream layout change as loudly as intended

The checksums are the digests GitHub reports for the release assets, not values recomputed locally — a truncated download hashes cleanly, and using one would produce a recipe that fails `do_fetch` with an error pointing at upstream rather than at this file.

### DIFF
--- a/recipes-containers/buildkit/buildkit_0.32.2.bb
+++ b/recipes-containers/buildkit/buildkit_0.32.2.bb
@@ -1,0 +1,91 @@
+SUMMARY = "BuildKit daemon, for devices acting as a wendy build host"
+DESCRIPTION = "buildkitd + buildctl, so `wendy run --build-host <device>` can \
+build container images on this device instead of a developer's laptop."
+HOMEPAGE = "https://github.com/moby/buildkit"
+LICENSE = "Apache-2.0"
+LIC_FILES_CHKSUM = "file://${COMMON_LICENSE_DIR}/Apache-2.0;md5=89aea4e17d99a7cacdbeed46a0096b10"
+
+# Upstream's own release binaries, pinned by version AND checksum for the same
+# reason wendyos-agent pins both: the version is part of SRC_URI and the checksum
+# is part of the do_fetch signature, so a bump re-runs the fetch instead of
+# serving a stale cached tarball.
+#
+# 0.32.2 is the version verified end to end on a Jetson AGX Thor and a DGX Spark
+# acting as build hosts. Bump version and BOTH checksums together.
+BUILDKIT_VERSION = "0.32.2"
+BUILDKIT_RELEASE_ARCH = "arm64"
+BUILDKIT_RELEASE_ARCH:x86-wendyos = "amd64"
+
+SRC_URI = "https://github.com/moby/buildkit/releases/download/v${BUILDKIT_VERSION}/buildkit-v${BUILDKIT_VERSION}.linux-${BUILDKIT_RELEASE_ARCH}.tar.gz;name=buildkit \
+           file://buildkit.service \
+           file://buildkitd.toml"
+
+SRC_URI[buildkit.sha256sum] = "${@d.getVar('BUILDKIT_SHA256_' + d.getVar('BUILDKIT_RELEASE_ARCH'))}"
+BUILDKIT_SHA256_arm64 = "9e8f46bf309ec0ab262967be5538a4dbe06be756a82621f98253933bac5dcf92"
+BUILDKIT_SHA256_amd64 = "2975d0f651ad96ba8b80b9992ae1f9a964f4408569af5b6dc36544165c3926af"
+
+S = "${UNPACKDIR}"
+
+inherit systemd
+
+# NOT auto-enabled. Running other people's builds is something a device opts
+# into: the agent gates it on its own marker file (build-host-enabled, set by
+# `wendy cloud device build-host enable`), and starting the daemon on every
+# device would spend memory on hosts that will never build. The unit is
+# installed and left disabled; enabling the role starts it.
+SYSTEMD_SERVICE:${PN} = "buildkit.service"
+SYSTEMD_AUTO_ENABLE:${PN} = "disable"
+
+COMPATIBLE_MACHINE = "^(?!qemuall).*$"
+
+do_install() {
+    # /opt, not /usr/local: a merged sysext driver add-on turns /usr into an
+    # ephemeral overlay, so anything installed under it is discarded on the next
+    # reboot (same reason wendyos-agent lives in /opt/wendyos/bin).
+    install -d ${D}/opt/buildkit/bin
+
+    # The tarball unpacks to bin/. Install every shipped binary rather than a
+    # hand-listed subset: buildkitd needs its runc and CNI helpers, and the
+    # buildkit-qemu-* binaries are what a future emulated-platform build would
+    # use. Find them rather than globbing a hard-coded path so an upstream
+    # layout change fails loudly here instead of silently shipping nothing.
+    BINDIR=$(find ${S} -type d -name bin | head -1)
+    if [ -z "${BINDIR}" ]; then
+        bbfatal "no bin/ directory in the unpacked buildkit release archive"
+    fi
+    for f in "${BINDIR}"/*; do
+        install -m 0755 "$f" ${D}/opt/buildkit/bin/
+    done
+    test -x ${D}/opt/buildkit/bin/buildkitd || bbfatal "buildkitd missing from the release archive"
+    test -x ${D}/opt/buildkit/bin/buildctl  || bbfatal "buildctl missing from the release archive"
+
+    # The agent execs a bare "buildctl" (build_service.go), so it must be on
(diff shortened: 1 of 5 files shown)


recipes-containers/buildkit/buildkit_0.32.2.bb
```diff
@@ -0,0 +1,91 @@
...
+
+# NOT auto-enabled. Running other people's builds is something a device opts
+# into: the agent gates it on its own marker file (build-host-enabled, set by
+# `wendy cloud device build-host enable`), and starting the daemon on every
+# device would spend memory on hosts that will never build. The unit is
+# installed and left disabled; enabling the role starts it.
+SYSTEMD_SERVICE:${PN} = "buildkit.service"
+SYSTEMD_AUTO_ENABLE:${PN} = "disable"
+
+COMPATIBLE_MACHINE = "^(?!qemuall).*$"
```
abularca, line 39: Is the intent to exclude `QEMU`? Because the `COMPATIBLE_MACHINE` regular expression doesn't.
`base.bbclass` loops over every `MACHINEOVERRIDES` entry and breaks on the first `re.match`. For `qemuarm64-wendyos`, that list contains `aarch64`, `armv8a`, `qemuarm64-wendyos` … — the first entry not literally beginning with "qemuall" satisfies the negative lookahead, so the recipe is compatible everywhere, including `QEMU`. 

**Proposal**:
Drop it as the `WENDYOS_BUILD_HOST` image gate already decides, or express the constraint that's actually real: 
`COMPATIBLE_HOST = "(aarch64|x86_64).*"`, 
since only `arm64` and `amd64` tarballs exist.

recipes-containers/buildkit/buildkit_0.32.2.bb
```diff
@@ -0,0 +1,91 @@
+SUMMARY = "BuildKit daemon, for devices acting as a wendy build host"
+DESCRIPTION = "buildkitd + buildctl, so `wendy run --build-host <device>` can \
+build container images on this device instead of a developer's laptop."
+HOMEPAGE = "https://github.com/moby/buildkit"
+LICENSE = "Apache-2.0"
```
abularca, line 5: I'm not so sure this license is the right option. 
`QEMU` is `GPL-2.0`-only. Either declare `Apache-2.0 & GPL-2.0`-only with matching LIC_FILES_CHKSUM, or don't install them.

recipes-containers/buildkit/buildkit_0.32.2.bb
```diff
@@ -0,0 +1,91 @@
...
+FILES:${PN} = "/opt/buildkit/bin/* \
+               ${bindir}/buildctl \
+               ${bindir}/buildkitd \
+               ${sysconfdir}/buildkit/buildkitd.toml \
+               ${systemd_system_unitdir}/* \
+               "
+
+# Prebuilt upstream binaries: no source to strip or split, and they are already
+# built for the target arch.
+INHIBIT_PACKAGE_STRIP = "1"
```
abularca, line 85: By default, `do_package` walks every ELF file under `${D}` and does two things to each one, in this order:
1. **Split**: 
Copies the debug sections out into /usr/lib/debug/<path> and puts them in an automatically created ${PN}-dbg package. Implemented in `meta/lib/oe/package.py`, guarded by `INHIBIT_PACKAGE_DEBUG_SPLIT`.
3. **Strip**: 
Runs the strip command on the copy that stays in ${PN} (guarded by `INHIBIT_PACKAGE_STRIP`). 

These lines switch both of those steps off:
```
83  # Prebuilt upstream binaries: no source to strip or split, and they are already
84  # built for the target arch.
85  INHIBIT_PACKAGE_STRIP = "1"
86  INHIBIT_PACKAGE_DEBUG_SPLIT = "1"
87  INHIBIT_SYSROOT_STRIP = "1"
```
so all 13 binaries go into the package byte-for-byte as they came out of upstream's tarball.

Proposed change:
```
--- a/recipes-containers/buildkit/buildkit_0.32.2.bb
+++ b/recipes-containers/buildkit/buildkit_0.32.2.bb
@@ -83,9 +83,10 @@
-# Prebuilt upstream binaries: no source to strip or split, and they are already
-# built for the target arch.
-INHIBIT_PACKAGE_STRIP = "1"
-INHIBIT_PACKAGE_DEBUG_SPLIT = "1"
-INHIBIT_SYSROOT_STRIP = "1"
+# Prebuilt upstream binaries, statically linked: there are no library
+# dependencies for the shlib/filedep scan to find, so those two scans are
+# switched off below. Stripping is NOT switched off -- upstream ships them
+# unstripped (file(1): "with debug_info, not stripped"), 67 MB of debug info
+# across the 13 binaries, and OE's normal strip + debug-split puts that in
+# buildkit-dbg, which no image installs.
 EXCLUDE_FROM_SHLIBS = "1"
 SKIP_FILEDEPS:${PN} = "1"
```

abularca (changes requested):

abularca:
Besides the above comments, this is what I would do overall:
1. Install two binaries instead of thirteen:
Keep buildkitd and buildctl, drop the rest. This saves space.
2. Use the runtime and plugins the image already has
- The image already has /usr/bin/runc (a symlink to crun). buildkit looks for runc on PATH and will find it.
- The image already has /opt/cni/bin with bridge, host-local, loopback and firewall. That is exactly the directory buildkit looks in by default.

    Both arrive with packagegroup-wendyos-container, which every board installs.
3. Add one line so builds get their own network
`networkMode = "bridge"`
4. Declare the two things bridge mode needs
`RDEPENDS:${PN} += "ca-certificates cni iptables"`
    In bridge mode, buildkitd refuses to start if the CNI plugins are missing.
5. Fix the unit
Change `Type=simple` to `Type=notify` - `buildkitd` tells systemd when its socket is actually ready, and without this a build fired immediately after enabling the role can fail to connect.

PR #242, opened 2026-08-24
sysext-apply: key the store by kernel and isolate add-on failures

kelenyibenjamin:
Fixes 3 problems in the driver add-on apply path, all found testing WDY-1925 on an rpi5 with agent: the script's exit status aggregated failures across every enabled add-on, so one broken add-on failed an unrelated install and the agent's rollback then deleted the good one; the store held one file per add-on name, so a rebuild for a new kernel overwrote the old one and an A/B rollback landed on a slot with no driver and neither a lock nor unit ordering stopped the boot unit and an agent-driven install from overlapping.

### DIFF
--- a/recipes-core/wendyos-sysext-apply/files/wendyos-sysext-apply.service
+++ b/recipes-core/wendyos-sysext-apply/files/wendyos-sysext-apply.service
@@ -4,6 +4,9 @@ Documentation=man:systemd-sysext(8)
 # The add-on store lives on /data. RequiresMountsFor ensures the mount is available
 # before this unit runs and fails the unit if /data cannot be mounted.
 RequiresMountsFor=/data
+# The agent migrates this store at startup, renaming the files this unit globs,
+# and both drive the same /usr merge.
+Before=wendyos-agent.service
 
 [Service]
 Type=oneshot
--- a/recipes-core/wendyos-sysext-apply/files/wendyos-sysext-apply.sh
+++ b/recipes-core/wendyos-sysext-apply/files/wendyos-sysext-apply.sh
@@ -10,6 +10,10 @@
 # as root. /data must be a trusted store.
 set -u
 
+# Optional: the add-on this run is for. The exit status then reports only that one,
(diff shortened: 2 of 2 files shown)


recipes-core/wendyos-agent/files/wendyos-agent.service
```diff
@@ -1,6 +1,8 @@
 [Unit]
 Description=Wendy Agent
-After=network.target dbus.service containerd.service gadget-setup.service
+# Ordered after the add-on merge: the agent migrates the driver store at startup,
+# renaming the very files that unit globs, and both drive the same /usr merge.
+After=network.target dbus.service containerd.service gadget-setup.service wendyos-sysext-apply.service
```
abularca: I might be wrong but I think the `After=` line is in wrong file. 
`wendyos-agent` is in `packagegroup-wendyos-base.bb`, so `wendyos-agent.service` ships on every board. On the other hand, `wendyos-sysext-apply.service` ships only when `WENDYOS_DRIVER_EXTENSIONS=1` and the machine is rpi. So on 10 of 11 boards the PR adds an ordering dependency on a unit that does not exist.
Thus, I would move the `After=` into the unit that only exists when the feature does:
```
--- a/recipes-core/wendyos-sysext-apply/files/wendyos-sysext-apply.service
+++ b/recipes-core/wendyos-sysext-apply/files/wendyos-sysext-apply.service
 RequiresMountsFor=/data
+# The agent migrates this store at startup, renaming the very files this unit
+# globs, and both drive the same /usr merge.
+Before=wendyos-agent.service
```
and revert `wendyos-agent.service` to its previous `After=` line.
Does it make sense?

kelenyibenjamin: good catch, and you are right 
the edge belongs on the unit that only exists with the feature
moved it to wendyos-sysext-apply.service as Before=wendyos-agent.service and reverted the agent unit as you suggested

abularca (approved):

PR #250, opened 2026-08-31
ci: drop download stamps whose payload the cache prune removed

kelenyibenjamin:
# Problem:
The daily cache prune (`find -mtime +30 -delete`) deleted `downloads/uninative/<sha>/x86_64-nativesdk-libc-5.2.tar.xz` but kept its `.done` stamp. `uninative.bbclass` gates the re-fetch solely on that stamp, so bitbake skips the download and then fails to untar a file that isn't there uninative silently disabled on every run since.

`Build jetson-agx-thor (nvme)` failed every run from 2026-08-31T00:01Z.

# Solution:
Drop any `downloads/**/*.done` whose payload is gone, so bitbake re-fetches.
The cache pool is shared, so the first job of a run repairs it for every branch.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -247,6 +247,17 @@ jobs:
             ln -sfn "$base/$d" "$GITHUB_WORKSPACE/$d"
             echo "linked $GITHUB_WORKSPACE/$d -> $base/$d"
           done
+          # The cache pool is shared and pruned on a timer we do not control, so
+          # validate it before trusting it. A fetch is gated purely on its .done
+          # stamp, so a stamp outliving its payload makes bitbake skip the
+          # download and then fail on the missing file, permanently.
+          # -e, not -f: git2/*.done stamps guard bare clone directories.
+          if [ -d "$base/downloads" ]; then
+            find "$base/downloads" -name '*.done' -type f | while IFS= read -r s; do
+              p="${s%.done}"
+              [ -e "$p" ] || { echo "repair: dropping orphaned stamp $s"; rm -f "$s"; }
+            done
+          fi
 
       # ---------- Build ----------
 


justsem (approved):
Approving — the diagnosis is right and the repair is correctly scoped for what it does. Confirmed the mechanism upstream: `uninative.bbclass` gates the whole fetch on `os.path.exists(tarballpath + ".done")` and then untars unconditionally, so a stamp outliving its payload does disable uninative permanently. Every other fetcher self-heals here (`fetch2.verify_donestamp()` removes a stamp whose payload is missing), which makes uninative the one bypass this repair actually covers. `-e` over `-f` is also right — git's `ud.localfile = ud.clonedir`, so `git2/*.done` really does guard a directory.

Security-wise it fits: the block runs as the build uid on a mount the launcher already gives the job `rw` (`ci` `spawn_test.go:43`), no sudo, nothing outside the tenant namespace, and `rm -f` is bounded to `*.done` so the worst case is a redundant download. Worth noting it's the first *destructive* operation on shared cross-branch state in a PR-runnable step — not a new capability (a PR can already write anything under `downloads/`), but the "PR jobs never mutate shared state, sstate/downloads are exempt because they're append-only" reasoning in the surrounding comments no longer holds as written.

Two things for after the merge:

1. **The root cause is ours, in `wendylabsinc/ci`** — the prune isn't "a timer we do not control", it's `prune: { max_age_days: 30 }` in `repositories/WendyOS-Builder/configuration.yaml`, one `find -type f -mtime +30 -delete` in `modules/platform/cache.nix`. bitbake deliberately keeps the last-use signal in the `.done` stamp (`update_stamp()` re-touches it on every use, "to show active use of the download") and never touches payloads again, so keyed on payload mtime the pruner evicts hot downloads and keeps their self-refreshing stamps. Fixed in wendylabsinc/ci#42: age off the stamp, evict the pair as a unit, plus a one-time repair of the stamps already orphaned. Once that's live this block is redundant and can be reverted.

2. **The expensive symptom is left open by the `-e` check.** `-type f` also descends into `downloads/git2/<repo>` bare mirrors: a repo that hasn't re-fetched in 30 days loses its packs while the directory *and* its stamp survive, so `[ -e "$p" ]` passes and the repair skips it. bitbake then fails `git rev-parse --absolute-git-dir`, `rmtree`s the clone and does a full re-clone (`fetch2/git.py:407-440`) — 5.7 GB for raspberrypi/linux, which looks like the recurring RPi `do_fetch` stall/exit-128 we keep retrying past. ci#42 holds those trees opaque.

Nits, none blocking:

- `find | while` is the last statement of a `set -euo pipefail` step, so any find diagnostic fails the step and kills the build before it starts. I confirmed that for an unreadable directory (exit 1, the loop's trailing statements never run); I tried 40 iterations to trigger it from concurrent deletion and couldn't, so it's unlikely in practice — but `|| true` on a repair path costs nothing.
- The comment should point at `repositories/WendyOS-Builder/configuration.yaml` rather than an uncontrolled timer.
- The repair only runs at job start; the daily timer can still fire inside a 240-minute build.

PR #255, opened 2026-09-07
 add Qualcomm Dragonwing IQ-8275 EVK support

kelenyibenjamin:
Adds the `iq-8275-evk-wendyos` machine on top of meta-qcom: A/B rootfs on UFS with wendyos-update, GRUB A/B boot, Wi-Fi and Ethernet, and CI that builds and publishes the qcomflash bundle so the board can be flashed over EDL with qdl.

Validated on hardware end to end, including an OTA install, slot switch and auto-commit.

 Refs: WDY-2846


abularca (commented):
Good PR :)

PR #257, opened 2026-09-09
 enable the Hexagon NPU on the Dragonwing IQ-8275

kelenyibenjamin:
Installs the Qualcomm AI stack into the image: the Hexagon DSP userspace +  the QAIRT/QNN

### DIFF
--- a/conf/distro/wendyos.conf
+++ b/conf/distro/wendyos.conf
@@ -169,6 +169,7 @@ require ${@'conf/distro/include/qemu-distro.inc' if 'qemuall' in d.getVar('MACHI
 require ${@'conf/distro/include/rpi-distro.inc'  if 'rpi'     in d.getVar('MACHINEOVERRIDES').split(':') else ''}
 require ${@'conf/distro/include/x86-distro.inc'  if 'x86-wendyos' in d.getVar('MACHINEOVERRIDES').split(':') else ''}
 require ${@'conf/distro/include/vm-distro.inc'   if 'vm-wendyos'  in d.getVar('MACHINEOVERRIDES').split(':') else ''}
+require ${@'conf/distro/include/qcom-distro.inc' if 'qcom-wendyos' in d.getVar('MACHINEOVERRIDES').split(':') else ''}
 
 # Driver add-ons (systemd-sysext): systemd's sysext support, the boot-time merge
 # service and the kernel devkit. Gated on WENDYOS_DRIVER_EXTENSIONS, so it must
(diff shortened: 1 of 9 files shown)


meta-qcom-extensions/recipes-ml/qairt/qairt-sdk_%.bbappend
```diff
@@ -0,0 +1,4 @@
+# A skel package is recommended for every Hexagon version supported. The SoC fixes which
+# one is usable and packagegroup-wendyos-qcom names it, so drop the recommendations
+# rather than refusing them individually.
+RRECOMMENDS:${PN} = ""
```
abularca: By using the `RRECOMMENDS:${PN} = ""` approach it permanently swallows anything upstream adds later and this is dangerous.
`RRECOMMENDS:${PN}:remove = "something..."` much better states the intent.

conf/distro/wendyos.conf
```diff
@@ -137,6 +137,22 @@ WENDYOS_DRIVER_EXTENSIONS := "${@d.getVar('WENDYOS_DRIVER_EXTENSIONS') if d.getV
 # may rely on.
 WENDYOS_SYSEXT_LEVEL ?= "1"
 
+# Hexagon DSP userspace selection. The signed DSP firmware authorises each Hexagon binary
+# by ELF segment hash, so this must name the build the shipped linux-firmware admits.
+# hexagon-dsp-binaries carries several; verify a change with its scripts/check.py.
+#
+# Distro-wide: hexagon-dsp-binaries is allarch and must not gain a machine-dependent
+# signature.
+WENDYOS_QCOM_DSP_BUILD ?= "DSP.AT.1.0.1-00201-LEMANS-2"
```
abularca: I would add these 3 new variables in `conf/distro/include/*-distro.inc` to use the same pattern like the other machines.

scripts/qcom-npu.test.sh
```diff
@@ -0,0 +1,109 @@
+#!/usr/bin/env bash
```
abularca, line 1: This script is located in the `recipes-*` folder rather than `files/`. It is harmless but it's a new placement convention, which is not a Yocto best practice. 
I'm not sure that `packagegroup` in the file name is usefull though :D

conf/distro/wendyos.conf
kelenyibenjamin: Done. 
Moved to `conf/distro/include/qcom-distro.inc`

scripts/qcom-npu.test.sh
kelenyibenjamin: Placed `packagegroup-wendyos-qcom.test.sh` under  `scripts/qcom-npu.test.sh`

conf/distro/wendyos.conf
kelenyibenjamin: agreed on the risk

instead of using a hardcoded `:remove` list, the code removes entries based on their name pattern

`:remove` list could cause problems later: `qairt-sdk` does not have a `qcs9100` path yet, so when IQ-9075 support is added upstream, a new `hexagon-vNN` entry could be added to the same list and accidentally get included again

the new python code removes anything matching `<PN>-hexagon-v<N>` and keeps all other entries. This also avoids hardcoding a specific board into a distro-wide bbappend.

abularca (approved):

PR #260, opened 2026-09-13
Detect the micro-USB cable instead of forcing host

abularca:
The port was pinned to host with its 5 V rail held on permanently. That socket is micro-AB and also takes an ordinary micro-B cable, so the board would push 5 V into whatever it was connected to.

The board already detects the cable -- the ID pin is wired to PMIC 0 and described in the device tree. Only CONFIG_USB_CONN_GPIO was missing, and that is off by accident: prune.config unsets ARCH_TEGRA, and the Tegra PHY was the only thing selecting it. With the driver built in, VBUS comes up only in host role. Qualcomm's own device tree for this board does the same. The refgen fix stays as it is what makes the controller probe at all.

Gadget networking stays broken until gadget-setup.sh selects its UDC explicitly: the port registers one again now, and a400000.usb sorts ahead of a600000.usb.

Refs: WDY-2846


kelenyibenjamin:
I fully agree with this change. Tested on HW, works fine.

kelenyibenjamin (approved):

PR #261, opened 2026-09-14
Enable existing driver add-on support on AGX Jetsons

thombles:
## Scope

Enable WendyOS's existing driver add-on machinery on AGX Thor (NVMe) and AGX
Orin (NVMe/eMMC).

- Opt the machines into the existing sysext runtime and kernel prerequisites.
- Produce matching kernel devkits for external module builds.
- Publish the Orin eMMC catalog identity alongside the generic Orin identity.
- Fix PR devkit lookup/publication and serialize PR driver indexing, so the
  existing add-on workflow can be exercised on these targets.

This PR does not introduce firmware downloads, new activation behaviour, Intel
backport build support, or a BE202 package. Ordinary existing add-ons such as
`wendyos-hello` are the intended platform-enablement example.

## Stack

1. This PR: enable existing driver support on Jetson.
2. #266 (`feat/driver-framework-extensions`): self-contained firmware bundling and
   declarative activation capabilities.
3. #262: Intel BE202 build adapter and hardware package.

Keep draft. This is a scope/history split of the existing stack, not a new
driver implementation.


.github/workflows/build.yml
```diff
@@ -1073,6 +1076,16 @@ jobs:
...
             # retries transient errors internally.
             "$RUNNER_TEMP/publisher" "${ARGS[@]}"
+            # Some installed images report a storage-specific board identity.
+            # Publish that identity as well, referencing these same uploaded
+            # bytes and this variant's devkit, without uploading another image.
+            alias=$(jq -r --arg key "$DEVICE/$STORAGE" '.[$key].manifest_alias // empty' \
+              "$GITHUB_WORKSPACE/wendyos/.github/device-artifacts.json")
+            if [[ -n "$alias" ]]; then
+              [[ "$alias" =~ ^[A-Za-z0-9._-]+$ ]] || { echo "invalid manifest alias" >&2; exit 1; }
+              jq --arg alias "$alias" '.device = $alias' "$ENTRY_FILE" \
```
kelenyibenjamin: .device is being rewritten with an identity that the device-keyed logic does not recognize. This creates a mismatch between the publisher and CLI sides, causing the install flow to fall back to a stale legacy path

possible fix: don't rewrite .device. Instead, add Aliases []string to ManifestEntry and have --apply-metadata write the same computed VersionMetadata for both the canonical device and its aliases

.github/workflows/drivers.yml
```diff
@@ -104,20 +106,26 @@ jobs:
...
-            # Prefer the requested version, else the newest version that has a
-            # devkit at all — a device that has never published one is skipped.
-            sel=$(jq -c --arg want "${WANT_VERSION:-}" '
-              [ .versions // {} | to_entries[]
-                | select(.value.devkit != null)
-                | {version: .key, devkit: .value.devkit,
-                   nightly: (.value.is_nightly // false)} ]
-              | (map(select(.version == $want)) | .[0]) // (sort_by(.version) | last)
-              // empty' <<<"$manifest")
+            urls=("${PUBLIC_BASE}/${PREFIX}manifests/${dev}.json")
```
kelenyibenjamin: drivers.yml:109-128 logs the same message whether the PR devkit or public devkit was selected

can you please log the selected URL/source and warn clearly on fallback. Otherwise, a failed OS leg can silently build add-ons against the old kernel, publish them under pr-<N>, and let the verify step pass against the wrong contract

kelenyibenjamin (changes requested):

.github/workflows/build.yml
thombles: Fixed, thanks

.github/workflows/drivers.yml
thombles: Fixed. We now log the source, make sure it it builds against the correct devkit, and warn on fallback.

scripts/select-driver-devkit.sh
```diff
@@ -0,0 +1,64 @@
...
+      | select(($devices | split(" ") | index($canonical)) != null) | .key' "$inventory")
+    [[ -n "$expected" ]] || exit 1
+    [[ -n "${WANT_VERSION:-}" ]] || fail "${dev}: current OS build version is missing"
+    shopt -s nullglob
+    entries=("${BUILT_ENTRIES:?}"/*.json)
+    [[ ${#entries[@]} -gt 0 ]] || fail "${dev}: no OS build metadata from this run; refusing public devkit fallback"
+    sel=$(jq -sc --arg dev "$dev" --arg version "$WANT_VERSION" \
+      --arg run "${GITHUB_RUN_ID:?}" --arg attempt "${GITHUB_RUN_ATTEMPT:?}" '
+      [.[] | select(.device == $dev or ((.aliases // []) | index($dev)) != null)
+        | select(.version == $version and .build_run_id == $run and .build_run_attempt == $attempt)
```
kelenyibenjamin: the selector requires `build_run_attempt` to match the current run:

`select(.version == $version and .build_run_id == $run and .build_run_attempt == $attempt)`

but artifacts survive re-runs, and untouched OS builds keep their original attempt number. So after Re-run failed jobs, GITHUB_RUN_ATTEMPT=2 while existing devkits still have attempt 1, causing all entries to be rejected

this forces Re-run all jobs, triggering a multi-hour Yocto rebuild just to recover from a failed publish

build_run_id already proves the entry belongs to the current run:

```
- select(.version == $version and .build_run_id == $run and .build_run_attempt == $attempt)
+ select(.version == $version and .build_run_id == $run)
```

you will need to keep --arg attempt since it is still used in the error message

tools/publisher/upload_and_manifest.go
kelenyibenjamin: this guard stops --swap recreating a legacy image path on a recovery-first T234 release

isRecoveryFirstT234 matches the literal strings "jetson-orin-nano" and "jetson-agx-orin". The alias identity jetson-agx-orin-emmc is not in that list

this didn't matter before, because the alias manifest carried install_mode: null

suggested fix:
```
-     if existingVersion.InstallMode == "recovery" && isRecoveryFirstT234(deviceType) {
+     if existingVersion.InstallMode == "recovery" {
```

kelenyibenjamin (approved):

PR #262, opened 2026-09-14
Add Intel BE202 Wi-Fi and Bluetooth driver for AGX Jetsons

thombles:
## Scope

Add the Intel BE202 Wi-Fi and Bluetooth package using the preceding framework
and Jetson-enablement PRs.

- Add the Intel-specific `backport-iwlwifi` adapter to the shared build script.
- Pin five backport modules: compat, cfg80211, mac80211, iwlwifi and iwlmld.
- Bundle checksum-pinned Wi-Fi/BT firmware and redistribution notices.
- Include the PCIe No-Snoop workaround scoped to BE202 behind the Thor root port.
- Declare the hardware match and activation operations in the package manifest.

Bluetooth uses the base OS's btusb/btintel modules with bundled firmware.
NetworkManager and wpa_supplicant are unchanged. This package does not configure
NAN, internet sharing or mesh routing, and performs no runtime firmware downloads.

Install a published bundle with `wendy device drivers install intel-be202`, or
use `--file <bundle.raw>` for a previously downloaded matching offline bundle.

## Stack

Third PR: based on #266 (`feat/driver-framework-extensions`), which is based on #261.
Keep draft. The combined implementation is unchanged by this stack reshuffle.


kelenyibenjamin:
build side looks clean and the patches ltgm. 3 things before merge:

1. wpa_supplicant.service doesn't exist on these boards, wifi isn't in DISTRO_FEATURES for tegra, so try-restart fails and every successful install reports failure
2. wlan0: can you confirm with ip link on a Thor with the card fitted? Wrong name fails silently after cfg80211 is already down
3. btusb: only the x86 packagegroup installs it. If it's missing on tegra, a BT module rolls back the whole Wi-Fi activation

thombles:
> * wpa_supplicant.service doesn't exist on these boards, wifi isn't in DISTRO_FEATURES for tegra, so try-restart fails and every successful install reports failure

This is working on device - `wifi` comes from the openembedded default DISTRO_FEATURES_DEFAULTS so it ends up in DISTRO_FEATURES. wpa_supplicant is installed and active.

> * wlan0: can you confirm with ip link on a Thor with the card fitted? Wrong name fails silently after cfg80211 is already down

Yes this tests fine on device with consistent name `wlan0`.

> * btusb: only the x86 packagegroup installs it. If it's missing on tegra, a BT module rolls back the whole Wi-Fi activation

In the case of Tegra it will always be present. With the current USB gadget configuration, usb-gadget-modules depends on kernel-modules and because we have `CONFIG_BT_HCIBTUSB=m` Yocto's build pulls that into the image.

kelenyibenjamin (approved):

PR #263, opened 2026-09-14
IQ-8275: restore Ethernet autonegotiation across supported speeds

thombles:
## Summary

Restore normal IQ-8275 Ethernet autonegotiation with a board-scoped kernel patch, based on `main`. No USB changes are included in this branch.

The QCA8081 switches its MAC-side interface between SGMII and 2500BASE-X according to the negotiated copper speed. Without `get_interfaces`, phylink validates only the firmware's initial `2500base-x` setting and filters out 10/100/1000BASE-T, so the board cannot link to a gigabit switch.

The patch supplies `get_interfaces` for the Qualcomm ETHQOS serial path, preserving the firmware interface and adding SGMII/2500BASE-X when validated by the SerDes. Existing runtime code applies the negotiated interface. No forced-speed configuration, PHY rate-matching workaround, or Ethernet DTS change is introduced. Builder applies the patch only to IQ-8275.

## Validation

- Strict Linux checkpatch: zero errors/warnings/checks; patch applies to the pinned Qualcomm kernel baseline.
- Full local Docker/Yocto image build, standard A/B OTA install, and two successful boots. Each boot automatically negotiated 1 Gb/s and obtained DHCP with the normal connection profile.
- Autonegotiation remained enabled throughout 1000 → 100 → 10 → 100 → 1000 Mb/s tests. Bidirectional TCP throughput was approximately 935 / 94 / 9 Mb/s respectively; matrix pings had zero loss. All supported modes were restored afterward.
- 2.5 Gb/s link confirmed with a direct Intel I226-V Windows peer: 2500 Mb/s/full, autonegotiation on, MAC-side interface changed from SGMII to 2500BASE-X without a reboot or configuration change.

The hardware image combined this Ethernet patch with #260 and its USB UDC-selection companion. 

Kernel upstream submission needs to be considered separately.


kelenyibenjamin:
Hi @thombles! went through the PR. LGTM but I have 1 small remark.

1. need a rebase
2.  log the SerDes switch failure. Your patch makes the interface genuinely flip on every speed change, so phy_set_mode_ext() is now a real re-lock that can return -ETIMEDOUT from any of four readiness polls in qcom_dwmac_sgmii_phy_calibrate(). ethqos_mac_finish_serdes() propagates it correctly, but stmmac_mac_finish() discards the glue's return and returns 0 unconditionally 
so phylink's major_config_failed path never arms, the link comes up carrying no traffic, and the only trace is a register-level "PLL Lock Status timed-out" with no netdev attached to it

```
@@ -719,9 +740,13 @@ static int ethqos_mac_finish_serdes(struct net_device *ndev, void *priv,
      qcom_ethqos_set_sgmii_loopback(ethqos, false);

      if (interface == PHY_INTERFACE_MODE_SGMII ||
-         interface == PHY_INTERFACE_MODE_2500BASEX)
+         interface == PHY_INTERFACE_MODE_2500BASEX) {
              ret = phy_set_mode_ext(ethqos->serdes_phy, PHY_MODE_ETHERNET,
                                     interface);
+             if (ret)
+                     netdev_err(ndev, "failed to switch SerDes to %s: %d\n",
+                                phy_modes(interface), ret);
+     }

```

thombles:
> 2\. log the SerDes switch failure

Good call. I thought it was a bit cleaner to propagate the error fully to where it will already be correctly logged and handled.

kelenyibenjamin (approved):

PR #264, opened 2026-09-14
IQ-8275: select USB0 explicitly for the USB gadget

thombles:
## Summary

Stacked on #260 (base: `feature/fix-dragonwing-usb-gadget`). This PR contains only the companion userspace UDC-selection fix.

Both IQ-8275 controllers register a UDC. The existing alphabetical selection chooses `a400000.usb` (ADB micro-B), moving the gadget away from USB0/USB-C. Select `a600000.usb` explicitly for this machine.

- Add machine configuration for the preferred UDC and install it in `/etc/default/gadget-setup`.
- Load the setting through the systemd service.
- Wait for the selected controller and report a timeout rather than silently binding another socket.
- Preserve the existing default selection for machines without an explicit preference; make the package machine-specific.

## Validation

- Installed the exact script/service with the generated machine setting on IQ-8275 running #260, then rebooted: USB0/USB-C enumerated and authenticated networking worked.
- Built and OTA-installed a combined test image; both subsequent boots automatically selected `a600000.usb`, with UDC state `configured` and usb0 carrier 1.
- Final USB IPv4 and IPv6 ping checks passed.
- Local selector tests passed for both controllers present, default selection, delayed selected controller, and missing selected controller; `bash -n` passed. Those local tests are not added by this PR.

No additional USB changes from the pre-publication review are included. This is independent of the wired Ethernet autonegotiation fix.


abularca:
This PR should be merged after PR #260 however, #260 has to be first tested on the hardware. Please, let's sync before merging.

thombles:
@abularca All good. #260 has been looking okay locally but I don't have an OTG dongle unfortunately so I can't confirm the main behaviour change right now.

abularca:
@thombles There is also #265 I've been working recently on, which addresses a more generic issue on the USB gadget setup. Sorry for the inconvenience. I'll ask Benjamin to help with the #260 and #256 testing and let you know. Thanks!

kelenyibenjamin:
Tested on HW with OTG adapter, works as expected.

kelenyibenjamin (approved):

PR #265, opened 2026-09-14
Fix USB gadget controller detection (using the board driver) and scope the unbind

abularca:
## Controller detection never worked outside Jetson

`gadget-setup.sh` chose the gadget's USB version by matching the module name in
`lsmod`, or the driver name against the UDC's name under `/sys/class/udc`.
Neither test can work. A UDC is named after its device-tree node, not its
driver, so it is `1000480000.usb` or `a600000.usb` and a `*dwc3*` match never
succeeds. A controller built into the kernel never appears in `lsmod` at all.

Only Jetson ever matched, because `tegra_xudc` happens to be a module there.

The fix reads the driver actually bound to the UDC, which works whether the
driver is built in or a module. The UDC wait moves ahead of detection, because
the controller cannot be identified before the UDC exists. Nothing between the
old and new positions affects when it appears: the modprobes in between load
gadget function modules, not controller drivers.

**What actually changes per board**

| Board | bcdUSB | Power | Tegra role switch |
| --- | --- | --- | --- |
| IQ-8275 | `0x0200` → **`0x0300`** | unchanged | n/a |
| RPi4 / RPi5 | unchanged | unchanged | n/a |
| Jetson | unchanged | unchanged | unchanged |

Only one board gains something. The IQ-8275 was advertising USB 2.0 on a
SuperSpeed-capable dwc3, and now advertises USB 3.0.

RPi was never wrong. Detection failed, the script fell through to its defaults,
and the defaults happened to be right on both counts -- `0x0200` is correct for
a USB 2.0 dwc2, and self-powered is correct for a Pi on its own PSU. The point
of the change there is that the right answer now comes from the right reason.

**This required removing the power-descriptor condition.** It was keyed on the
dwc2 controller, standing in for an OTG board powered from the host. It never
fired, because dwc2 is built in on RPi (`CONFIG_USB_DWC2=y` in both
`bcm2711_defconfig` and `bcm2712_defconfig`), so every board got self-powered by
fall-through. Repairing detection would have switched it on for the first time
and told hosts an RPi draws 500 mA from the bus, which is false for a board on
its own supply. How a board is powered is a property of the board, not of its
USB controller, and every board running this gadget has its own supply.

## The unbind service tore down gadgets it did not own

The udev rule fires on the removal of *any* udc and cannot say which one left,
so the service cleared the gadget's binding unconditionally. That is harmless
while a board has one peripheral-capable socket, and wrong as soon as it has
two. A dual-role port drops its UDC every time it switches to host mode, and the
gadget on the other controller would go down with it.

* move the decision into a helper that compares the bound controller against
  what is still present
* no board has to name its UDC in a shared rule
* the legitimate case -- our own controller disappearing -- still unbinds

This changes nothing today. `usb_2` is still pinned to host mode on main, so it
registers no UDC and the teardown cannot happen yet. The guard matters the
moment #260 lands and makes that port dual-role.

## Merge order

`#265` → `#260` → `#264`, and the last two should not sit far apart.

This PR shares no files with #260, so the two are independent. #260 is what
removes the host-mode pin on `usb_2`. From the moment it merges until #264
follows, the IQ-8275 has two UDCs and `ls | head -n1` picks `a400000.usb` -- the
micro-AB socket -- so the gadget binds to the wrong port. #264 is the only thing
that closes that window.

#264 also needs a rebase: it edits the UDC wait loop that this PR relocates to
the top of the script. Its `GADGET_UDC` selection has to go into the loop at its
new position, not the old one. A "keep the moved block" conflict resolution
drops that fix silently.

## Testing

The unbind helper was exercised against a fake sysfs over the six reachable
states, under dash, busybox sh and bash. Kernel claims in the comments were
checked against source. Neither change has been run on hardware yet.


recipes-core/gadget-setup/files/99-usb-gadget-udc.rules
```diff
@@ -1,3 +1,8 @@
 # /etc/udev/rules.d/99-usb-gadget-udc.rules
+#
+# This fires for the removal of any udc, and does not say which one went away.
+# Boards with two peripheral-capable sockets remove a UDC in normal use, so the
+# service it starts checks whether the bound controller is the one that left
+# before it touches the binding. Do not make that check unconditional again.
 
 SUBSYSTEM=="udc", ACTION=="remove", TAG+="systemd", ENV{SYSTEMD_WANTS}+="wendyos-usbgadget-unbind.service"
```
kelenyibenjamin, line 8: even if this is on main, but systemd ignores ENV{SYSTEMD_WANTS} on ACTION=="remove", so wendyos-usbgadget-unbind.service never starts and I think the entire unbind path this PR reworks is unreachable dead code

recipes-core/gadget-setup/files/usbgadget-unbind
```diff
@@ -0,0 +1,31 @@
...
+[ -e "$GADGET/UDC" ] || exit 0
+
+UDC=$(tr -d ' \t\n' < "$GADGET/UDC" 2>/dev/null) || exit 0
+[ -n "$UDC" ] || exit 0
+
+# The controller we are bound to is still here, so this removal was someone
+# else's. Leave the binding alone.
+[ -e "/sys/class/udc/$UDC" ] && exit 0
+
+echo "" > "$GADGET/UDC" 2>/dev/null || true
```
kelenyibenjamin, line 28: clearing $GADGET/UDC unregisters the configfs gadget driver and destroys the kernel's own automatic re-bind nothing in the image ever re-binds. Is this the wanted behavior?

recipes-core/gadget-setup/files/99-usb-gadget-udc.rules
abularca: You are right!

recipes-core/gadget-setup/files/usbgadget-unbind
abularca: Correct as well, and it goes further than the re-bind being lost: the kernel already handles this case on its own.
The unbind service should be taken out but is is safe to remove it in a dedicated PR.

kelenyibenjamin:
I tested on rpi5 and dragonwing IQ-8275.


**dragonwing IQ-8275**

* USB-C gadget correctly binds to `a600000.usb` when the controller is selected
* Fixed teardown: gadget now fully removes without `rmdir` errors
* `gadget-setup` restart succeeds and `usb0` stays up

**rpi5**

* `dwc2` is now detected correctly on cold boot
* USB power descriptor is correct (`0xc0`, `2`)
* `gadget-setup` succeeds and `usb0` stays up

kelenyibenjamin (approved):

PR #266, opened 2026-09-14
Extend driver bundles with firmware and declarative activation

thombles:
## Scope

Extend the existing driver framework to support self-contained firmware payloads
and controlled replacement of an already-loaded driver stack.

- Fetch pinned, checksum-verified firmware at BUILD TIME and embed it in the
  `.raw` bundle. Installation and activation do not download firmware.
- Add manifest device allowlists for CI targeting.
- Add declarative hardware matching, ordered module replacement/reload, service
  restart and restoration of the active NetworkManager Wi-Fi profile.
- Stop replacement when an existing module cannot unload.
- Add targeting and activation regression tests; retain ordinary add-on behaviour.

The OS implements activation; packages declare supported operations rather than
shipping arbitrary post-install scripts. A previously downloaded matching bundle
can be installed with `wendy device drivers install <name> --file <bundle.raw>`.

No Intel-specific build adapter, firmware pins, hardware IDs or BE202 package
are introduced here. NetworkManager profile restoration is explicitly
networking-aware framework functionality.

## Stack

Based on #261 (Jetson platform enablement). #262 sits on top and supplies the
Intel backport build adapter and BE202 package. Keep draft.


recipes-core/wendyos-sysext-apply/files/wendyos-sysext-apply.sh
```diff
@@ -131,33 +399,148 @@ for link in "$RUNDIR"/*.raw; do
...
+    activate_now=0
+    if [ -z "$SUBJECT" ] || [ "$SUBJECT" = "$name" ]; then activate_now=1; fi
+    removed="$PLAN/$name.removed"
+    inserted="$PLAN/$name.inserted"
+    restore_state="$PLAN/$name.restore"
+    : > "$removed"
+    : > "$inserted"
+    : > "$restore_state"
+
+    if [ -f "$activation" ] && [ "$activate_now" = 1 ]; then
```
kelenyibenjamin: wifi is never restored on any rollback path
capture_wifi_state runs before the unload, because after it the interface is gone

proposed fix: hoist the call out of the if at :533-542, or add a finish_package() invoked from :480, :492, :529 and :541

recipes-core/wendyos-sysext-apply/files/wendyos-sysext-apply.sh
```diff
@@ -92,16 +98,18 @@ if ! systemd-sysext --mutable=ephemeral --always-refresh=yes refresh; then
     exit 1
 fi
 
+PLAN=$(mktemp -d "/run/wendyos-driver-plan.XXXXXX") || exit 1
```
kelenyibenjamin: SIGTERM deletes the rollback state, then the script reports a clean rollback

The handler never calls exit, so on signals 1/2/15 it deletes $PLAN and execution resumes. Every recovery function then short-circuits to success:

```
hide_payload() {
exposed=$1
[ -f "$exposed" ] || return 0  
[ -s "$modules" ] || return 0 
```

wendyos-sysext-apply.service is Type=oneshot with no TimeoutStartSec=, so systemd's 90 s default applies — and restore_wifi_state alone can burn udevadm settle --timeout=10 (:360) plus 30 × sleep 1 (:364-378) per interface before reaching nmcli connection up (:379), which has no -w. The unit is also Before=wendyos-agent.service, so this blocks the agent.

proposed fix: trap 'rc=$?; rm -rf "$PLAN"; exit $rc' 0 plus a separate signal trap that logs "reboot required" and exits; add explicit TimeoutStartSec=; pass nmcli -w <n>.

recipes-core/wendyos-sysext-apply/files/wendyos-sysext-apply.sh
```diff
@@ -131,33 +399,148 @@ for link in "$RUNDIR"/*.raw; do
...
+                    normalized=$(printf '%s' "$value" | tr 'A-F' 'a-f')
+                    pci_device_present "$normalized" && pci_matched=1
+                    ;;
+                replace|reload) [ -z "${extra:-}" ] && valid_module_name "${value:-}" || activation_bad=1 ;;
+                restore-wifi-connection) [ -z "${extra:-}" ] && valid_interface_name "${value:-}" || activation_bad=1 ;;
+                restart-service) [ -z "${extra:-}" ] && valid_service_name "${value:-}" || activation_bad=1 ;;
+                *) activation_bad=1 ;;
+            esac
+        done < "$activation"
+        if [ "$activation_bad" = 1 ]; then
```
kelenyibenjamin: wendy device drivers install reports success having installed nothing

the agent has no independent notion of success WendyOS/go/internal/agent/services/driver_service.go:~635-648

out, err := cmd.CombinedOutput()
if err != nil { return fmt.Errorf("%s: %v (%s)", s.applyScript, err, ...) }

so installing onto a board whose PCI ID doesn't match, or a package whose module is already claimed by an alphabetically-earlier one, returns clean success for a permanent no-op

recipes-core/wendyos-sysext-apply/files/wendyos-sysext-apply.sh
```diff
@@ -112,16 +120,276 @@ resolves() {
...
+            *[!A-Za-z0-9_.-]*) echo "wendyos-sysext-apply: bad module name '$mod'" >&2; return 1 ;;
+        esac
+        load_one "$mod" "$inserted" || return 1
+    done < "$conf"
+}
+
+capture_wifi_state() {
+    activation=$1
+    restore_state=$2
+    while read -r directive interface extra || [ -n "${directive:-}${interface:-}${extra:-}" ]; do
```
kelenyibenjamin: restore-wifi-connection silently does nothing

hen restore_wifi_state can't tell "captured nothing" from "nothing to restore":

[ -s "$restore_state" ] || return 0

and a failed activation is an echo only rc untouched:

nmcli connection up uuid "$connection_uuid" ifname "$interface" ||
echo "wendyos-sysext-apply: could not restore $interface connection" >&2

recipes-core/wendyos-sysext-apply/files/wendyos-sysext-apply.sh
```diff
@@ -131,33 +399,148 @@ for link in "$RUNDIR"/*.raw; do
...
+                echo "wendyos-sysext-apply: replacement aborted for $name; rollback incomplete, reboot required" >&2
+            fi
+            rc=$((rc|fail))
+            continue
+        fi
+    fi
+
+    if ! expose_payload "$name" "$payload" || ! depmod -a "$KVER"; then
+        echo "wendyos-sysext-apply: could not expose payload for $name" >&2
+        hide_payload "$PLAN/$name.exposed" ||
```
kelenyibenjamin: depmod failures inside both rollbacks are discarded, under a "rolled back" message

depmod -a "$KVER" >/dev/null 2>&1 || true    <----- swallowed

recipes-core/wendyos-sysext-apply/files/wendyos-sysext-apply.sh
```diff
@@ -131,33 +399,148 @@ for link in "$RUNDIR"/*.raw; do
...
+            [ "${directive:-}" = reload ] || continue
+            if ! load_one "$mod" "$inserted"; then package_failed=1; break; fi
+        done < "$activation"
+    fi
+
+    if [ "$package_failed" != 0 ]; then
+        rollback_incomplete=0
+        rollback_modules "$inserted" -r || rollback_incomplete=1
+        hide_payload "$PLAN/$name.exposed" || rollback_incomplete=1
+        depmod -a "$KVER" >/dev/null 2>&1 || true
```
kelenyibenjamin: depmod -a "$KVER" >/dev/null 2>&1 || true <_----- swallowed

kelenyibenjamin (changes requested):

recipes-core/wendyos-sysext-apply/files/wendyos-sysext-apply.sh
thombles: Thanks for all these - I think I introduced some bugs in a late change after I'd run most of my checks.

This is fixed. Every failure after unload uses the shared fallback path so the wifi connection is restored.

recipes-core/wendyos-sysext-apply/files/wendyos-sysext-apply.sh
thombles: Thanks, signal handling now exits and reports uncertain state. I also added a service timeout.

recipes-core/wendyos-sysext-apply/files/wendyos-sysext-apply.sh
thombles: Thanks, an explicit install returns an error if the hardware is absent or if there's a module collision.

recipes-core/wendyos-sysext-apply/files/wendyos-sysext-apply.sh
thombles: Fixed. Capture errors stop activation before unload, and confirmed absence of active connection is handled explicitly. Reconnect failures now fail activation and trigger fallback.

recipes-core/wendyos-sysext-apply/files/wendyos-sysext-apply.sh
thombles: Fixed. Incomplete recovery is now reported if depmod fails

recipes-core/wendyos-sysext-apply/files/wendyos-sysext-apply.sh
thombles: Fixed

recipes-core/wendyos-sysext-apply/files/wendyos-sysext-apply.sh
```diff
@@ -112,52 +194,563 @@ resolves() {
...
+        restore_previous "$(basename "$previous")" || echo "wendyos-sysext-apply: could not restore previous exposure" >&2
+    done
+    depmod -a "$KVER" || echo "wendyos-sysext-apply: could not rebuild previous module index" >&2
+    exit 1
+fi
+
+# Globs follow LC_ALL=C, making the first-wins policy stable across boots. Claims
+# belong to successfully activated candidates or restored previous payloads.
 rc=0
 for link in "$RUNDIR"/*.raw; do
```
kelenyibenjamin: when --subject is provided but no matching .raw file exists, the loop never handles it, so rc remains 0
the agent treats this as success because driver_service.go:~636-648 only checks the script's exit code

this can hide real failures such as a failed /data write, wrong kernel bucket, name mismatch, or read-only /data.

proposed fix:
```
 rc=0
+subject_seen=0
 for link in "$RUNDIR"/*.raw; do
     [ -e "$link" ] || continue
     name=$(basename "$link" .raw)
+    [ "$SUBJECT" != "$name" ] || subject_seen=1
     ...
 done

+if [ -n "$SUBJECT" ] && [ "$subject_seen" = 0 ]; then
+    echo "wendyos-sysext-apply: no enabled image named '$SUBJECT' for kernel $KVER" >&2
+    rc=1
+fi
 udevadm trigger >/dev/null 2>&1 || true
 exit "$rc"
```

also apply the same [ -n "$SUBJECT" ] check to the early exits at :65 and :87, so a targeted install into an empty store cannot incorrectly report success.

recipes-core/wendyos-sysext-apply/files/wendyos-sysext-apply.sh
```diff
@@ -85,23 +92,98 @@ if ! has_links; then
...
+)
+
+if ! snapshot_previous; then
+    echo "wendyos-sysext-apply: cannot preserve previous payloads; refusing refresh" >&2
+    exit 1
+fi
+
 # --always-refresh: reinstalling a driver reuses the image filename, which on its own looks
 # like no change.
 if ! systemd-sysext --mutable=ephemeral --always-refresh=yes refresh; then
```
kelenyibenjamin: removes the previous ephemeral layer on every run. snapshot_previous saves the old payload so it can be restored on failure, but the PCI-skip and collision paths currently just continue, leaving the previous payload deleted until the next boot

this is especially risky for devices that enumerate late: the driver may keep running until the next request_firmware() or reprobe, then fail with no clear link to the install

suggested fix: 

```
 if [ "$pci_declared" = 1 ] && [ "$pci_matched" = 0 ]; then
-    echo "wendyos-sysext-apply: $name has no matching PCI device — skipping" >&2
+    echo "wendyos-sysext-apply: $name has no matching PCI device — not applying" >&2
+    restore_previous "$name" && depmod -a "$KVER" ||
+        echo "wendyos-sysext-apply: could not restore previous payload for $name" >&2
     [ "$SUBJECT" != "$name" ] || rc=1
     continue
 fi

 if payload_collision "$name"; then
+    restore_previous "$name" && depmod -a "$KVER" ||
+        echo "wendyos-sysext-apply: could not restore previous payload for $name" >&2
     [ "$SUBJECT" != "$name" ] || rc=1
     continue
 fi
```

kelenyibenjamin:
Sorry Tom, I missed checking a few things in my first review iteration
I went through them again now and found a few additional issues

kelenyibenjamin:
#PR261, #PR262 -> approved

PR #272, opened 2026-09-16
publisher: batch master manifest writes and retry rate limits

thombles:
PR publishing and latest promotion rewrite the same GCS master manifest once per device. A ten-device run can exceed the object's mutation rate limit and fail with HTTP 429, as seen in #270, after all image builds have passed.

Add a batch publishing command and use it in both workflows. Eligible device pointers are applied with one read and one write on the successful path. The release driver checks still hold back devices missing required add-ons; unchanged devices, firmware entries, and the opposite release channel are preserved. Storage variants are deduplicated, conflicting rows are rejected before accessing storage, and an empty batch is a no-op.

Retry the complete transaction on rate limits, generation conflicts, and other transient storage errors with bounded, cancellable backoff. Generation checks preserve intervening updates, and only a genuine not-found response permits creating a new manifest. The existing single-device command uses the same implementation.

Validation:
- `go test -race ./...` and `go vet ./...` passed.
- Storage-client tests cover ten-device single writes, PR namespaces, channel/firmware preservation, 412/429/503 retries, concurrent writers, validation, cancellation, and permanent errors.
- Five tests execute the actual workflow scripts in Linux with cloud commands stubbed, including partial/all driver holdbacks and malformed metadata.
- Actionlint passed with the repository's custom runner labels configured; CLI validation and empty-batch smoke tests passed.

Storage behavior was tested against a local HTTP server using the real GCS client. Live publishing validation will come from this PR's CI run.


thombles:
CI-only change, appears to be working and I need it for another PR. Merging.

thombles:
On the HIGH security finding: this was pre-existing workflow behaviour that doesn't leak secrets to Action logs, although the token is temporarily visible within the CI runner in the process arguments. When we switch to Application Default Credentials (#229) this will no longer be a concern.

PR #273, opened 2026-09-17
dragonwing: add Qualcomm Dragonwing IQ-9075 EVK support

kelenyibenjamin:
Adds the Qualcomm Dragonwing IQ-9075 EVK (QCS9075 / SA8775P, board codename "lemans") as a supported board.

- `iq-9075-evk-wendyos` machine + `dragonwing-iq9075` board template, wrapping  meta-qcom's `iq-9075-evk`. Names only what differs: the device tree, the DSP triple (`iq9075-evk` / `v73` / `sa8775p`) and the board id.
- `conf/machine/include/dragonwing-common-wendyos.inc` holds the settings both boards share, following the existing `raspberrypi-common-wendyos.inc` pattern.
- CI builds and publishes the board alongside the 8275; README updated.


tools/publisher/upload_and_manifest.go
```diff
@@ -3617,6 +3783,10 @@ func swapImageFile(ctx context.Context, bucket *storage.BucketHandle, prefix, de
 		OTAUpdateChecksum:  existingVersion.OTAUpdateChecksum,
 		OTAUpdateSizeBytes: existingVersion.OTAUpdateSizeBytes,
 
+		QcomflashPath:      existingVersion.QcomflashPath,
```
thombles: Watch out that "Path" is always set up at L3715 so if we are relying on the empty-Path behaviour to prevent treating IQ-9075 flash bundles as a generic disk image then that will have to be preserved here

tools/publisher/upload_and_manifest.go
thombles: I believe this is going to trigger if we try to promote an IQ-9075 artifact since the generic path isn't set - probably the L3288 condition needs to be updated to handle EDL bundles

thombles (approved):
Two small publishing bugs to address, otherwise seems good

tools/publisher/upload_and_manifest.go
kelenyibenjamin: Correct, promoting a 9075 would have stopped here with "Version has no primary image path".

The check now accepts an EDL bundle as a primary artifact too, so a version is only rejected when it really has nothing to copy. I pulled it out into a small hasPromotableImage helper so it could be unit tested.

tools/publisher/upload_and_manifest.go
kelenyibenjamin: Right, Path was being set unconditionally here, which would have undone the empty-path behaviour for the 9075.

The swap now calls applyEDLFlashBundle, so the 9075 keeps its path empty and the 8275 keeps its path as before. I also made that helper clear the legacy fields itself instead of trusting callers not to set them.

kelenyibenjamin:
Thanks @thombles!
Both fixed, good catches!

